PR #2, opened 2015-07-16
Set LOGROTATE_COMPRESSED_FILENAME environment when compressing file.

zimathias:
The compress program doesn't know the name of the file being compressed. We're
using a custom wrapper for compressing files which needs to know the name of
the compressed file.

We could pass the name of the compressed file in the
logrotate configuration itself using compressoptions:

    /var/log/foo.log {
      
      compresscmd my_wrapper
      compressoptions /var/log/foo.log

    }

However this doesn't work with wildcard filename:

    /var/log/*log {
    
      compresscmd my_wrapper
      compressoptions ???
  
    }

given that wildcard filenames are only known when the compress command is
called.

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -524,6 +524,7 @@ static int removeLogFile(char *name, struct logInfo *log)
 static int compressLogFile(char *name, struct logInfo *log, struct stat *sb)
 {
     char *compressedName;
+    char *envInFilename;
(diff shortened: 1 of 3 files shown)


hanzz:
Sorry for late reaction. I've somehow missed this pull request.

PR #16, opened 2016-01-18
Fix broken file caused by full disk

XufengZhang:
This fix tries to handle this scenario:
When the disk is full, logrotate will die ungracefully and leave some
temporary files. After free some space in the disk, logrotate should
handle the previous erroneous state, instead of exiting with "File exists"
error, otherwise, logrotate can't rotate current logs normally and this
will make the disk full finally.

The solution is to rename the previously broken file with
"_broken_$timestamp" suffix append and retry one time to create the
output file again.

This is a long time known bug(Debian bug #693985).

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -425,6 +425,36 @@ int createOutputFile(char *fileName, int flags, struct stat *sb, acl_type acl, i
     return fd;
 }
 
+/*
+ * This function tries to handle this scenario:
+ * When the disk is full, logrotate will die ungracefully and leave some
+ * temporary files. After free some space in the disk, logrotate should
+ * handle the previous erroneous state, instead of exiting with "File exists"
+ * error, otherwise, logrotate can't rotate current logs normally and this
+ * will make the disk full finally.
+ * The solution is to rename the previously broken file with
+ * "_broken_$timestamp" suffix append and try to create this output file again.
+ */
+int retryCreateOutputFile(char *brokenFileName, int flags, struct stat *sb,
+			  acl_type acl, int force_mode)
+{
+    char *renamedFile;
+    char curr_time[14];
+    struct tm now = *localtime(&nowSecs);
+
+    message(MESS_DEBUG, "Rename broken file: %s\n", brokenFileName);
+    renamedFile = (char *)malloc(strlen(brokenFileName) + 24);
+    strftime(curr_time, sizeof(curr_time), "%m%d%H%M%Y", &now);
+    sprintf(renamedFile, "%s_%s_%s", brokenFileName, "broken", curr_time);
+    if (rename(brokenFileName, renamedFile)) {
+	message(MESS_ERROR, "failed to rename %s to %s: %s\n",
+		brokenFileName, renamedFile, strerror(errno));
+    }
+    free(renamedFile);
+
+    return createOutputFile(brokenFileName, flags, sb, acl, force_mode);
+}
+
 #define DIGITS 10
 
 /* unlink, but try to call shred from GNU fileutils */
@@ -569,7 +599,16 @@ static int compressLogFile(char *name, struct logInfo *log, struct stat *sb)
 		prev_acl = NULL;
 	}
 #endif
-    if (outFile < 0) {
+    if (outFile < 0 && errno == EEXIST) {
+	/* Handle previsouly erroneous state caused by disk full:
+	 * rename the existed file and try to create it again */
+	outFile = retryCreateOutputFile(compressedName, O_RDWR | O_CREAT,
+					sb, prev_acl, 0);
+	if (outFile < 0) {
+		close(inFile);
+		return 1;
+	}
+    } else if (outFile < 0) {
 	close(inFile);
 	return 1;
     }
@@ -935,7 +974,16 @@ static int copyTruncate(char *currLog, char *saveLog, struct stat *sb,
 		prev_acl = NULL;
 	}
 #endif
-	if (fdsave < 0) {
+	if (fdsave < 0 && errno == EEXIST) {
+		/* Handle previsouly erroneous state caused by disk full:
+		 * rename the existed file and try to create it again */
+		fdsave = retryCreateOutputFile(saveLog, O_WRONLY | O_CREAT,
+					       sb, prev_acl, 0);
+		if (fdsave < 0) {
+			close(fdcurr);
+			return 1;
+		}
+	} else if (fdsave < 0) {
 	    close(fdcurr);
 	    return 1;
 	}


hanzz:
Can you give me an example of files created by logrotate in case of the full disk? Ideally with the logrotate configuration for rotating them. I think it would be better to skip such rotation (or exit with an error) in case logrotate is not able to rotate the file.

I can imagine it can happen in case of "copytruncate" or "copy", but want to know what was causing that particular error for you.

XufengZhang:
In our product environment, we have met two kinds of cases, take irqbalance.log as an example:
1. Without "delaycompress" option
irqbalance.log {
    missingok
    size=20M
    rotate 7
    copytruncate
    compress
    su root admin
}
irqbalance.log becomes very large after some time:
# ls -lh irq*
-rw-r--r--  1 root admin 7.3G Oct  8 22:53 irqbalance.log
-rw-r--r--  1 root admin 2.1M Aug 25 09:45 irqbalance.log.1
-rw-r--r--  1 root admin 485K Aug 25 06:44 irqbalance.log.8.gz

With verbose output printed, we can find logrotate can't rotate irqbalance.log:
# logrotate -v /etc/logrotate.d/irqbalance.log
......
copying /../irqbalance.log to /../irqbalance.log.1
error: error creating output file /../irqbalance.log.1: File exists
switching euid to 0 and egid to 0

And after analysis, we know irqbalance.log.1 exist because of disk full at that time.

2. With "delaycompress" option
irqbalance.log {
    missingok
    size=20M
    rotate 7
    copytruncate
    compress
    delaycompress
    su root admin
}

irqbalance.log enters another error state and can't be rotated:
# ls -l irq*
-rw-rw-rw-  1 root admin  61653981 Jan 15 06:20 irqbalance.log
-rw-rw-rw-  1 root admin  35890420 Jan 13 00:45 irqbalance.log.1
-rw-rw-rw-  1 root admin         0 Jan 14 00:46 irqbalance.log.1.gz
-rw-rw-rw-  1 root admin   3784732 Jan 12 00:45 irqbalance.log.2.gz
-rw-rw-rw-  1 root admin   3718552 Jan 11 00:45 irqbalance.log.3.gz

# logrotate -v /etc/logrotate.d/irqbalance.log
......
compressing log with: /bin/gzip
error: error creating output file /../irqbalance.log.1.gz: File exists

And also after analysis, we found this zero-sized irqbalance.log.1.gz was caused by full disk.

So My fix is trying to resolve the above two scenarios.

XufengZhang:
Please check this comment, the above one seems in a mess, sorry.
In our product environment, we have met two kinds of cases, take irqbalance.log as an example:
 1. Without "delaycompress" option
 irqbalance.log {
 missingok
 size=20M
 rotate 7
 copytruncate
 compress
 su root admin
 }
 irqbalance.log becomes very large after some time:

\> ls -lh irq*
-rw-r--r-- 1 root admin 7.3G Oct 8 22:53 irqbalance.log
 -rw-r--r-- 1 root admin 2.1M Aug 25 09:45 irqbalance.log.1
 -rw-r--r-- 1 root admin 485K Aug 25 06:44 irqbalance.log.8.gz

With verbose output printed, we can find logrotate can't rotate irqbalance.log:
\> logrotate -v /etc/logrotate.d/irqbalance.log
......
 copying /../irqbalance.log to /../irqbalance.log.1
 error: error creating output file /../irqbalance.log.1: File exists
 switching euid to 0 and egid to 0

And after analysis, we know irqbalance.log.1 exist because of disk full at that time.

2. With "delaycompress" option 
irqbalance.log {
 missingok
 size=20M
 rotate 7
 copytruncate
 compress
 delaycompress
 su root admin
 }

irqbalance.log enters another error state and can't be rotated:
\> ls -l irq*
-rw-rw-rw- 1 root admin 61653981 Jan 15 06:20 irqbalance.log
 -rw-rw-rw- 1 root admin 35890420 Jan 13 00:45 irqbalance.log.1
 -rw-rw-rw- 1 root admin 0 Jan 14 00:46 irqbalance.log.1.gz
 -rw-rw-rw- 1 root admin 3784732 Jan 12 00:45 irqbalance.log.2.gz
 -rw-rw-rw- 1 root admin 3718552 Jan 11 00:45 irqbalance.log.3.gz

\> logrotate -v /etc/logrotate.d/irqbalance.log
......
 compressing log with: /bin/gzip
 error: error creating output file /../irqbalance.log.1.gz: File exists

And also after analysis, we found this zero-sized irqbalance.log.1.gz was caused by full disk.

So My fix is trying to resolve the above two scenarios.

XufengZhang:
BTW, we don't have the error log for logrotate while the disk is full.
And this was also ever reported by other people:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=693985

hanzz:
I think this could be fixed by following simple patch without any side effects. I will think about it twice before committing it, bus after quick look, it should work as expected:

```
diff --git a/logrotate.c b/logrotate.c
index 974b588..a8642a0 100644
--- a/logrotate.c
+++ b/logrotate.c
@@ -621,6 +621,7 @@ static int compressLogFile(char *name, struct logInfo *log, struct stat *sb)
     if (!WIFEXITED(status) || WEXITSTATUS(status)) {
        message(MESS_ERROR, "failed to compress log %s\n", name);
        close(inFile);
+       unlink(compressedName);
        return 1;
     }
 
@@ -943,6 +944,9 @@ static int copyTruncate(char *currLog, char *saveLog, struct stat *sb,
        if (sparse_copy(fdcurr, fdsave, sb, saveLog, currLog) != 1) {
                close(fdcurr);
                close(fdsave);
+               message(MESS_ERROR, "error copying %s to %s: %s\n", currLog,
+                       saveLog, strerror(errno));
+               unlink(saveLog);
                return 1;
        }
     }
```

We have just two cases where it is likely to get out of space:

1) "copytruncate" - In this case, logrotate at first copies the log to log.1 and then truncates log. If we get out of space during "cp log log.1", we can just remove "log.1", because all the data is still in the "log" and stop rotation of that file. This case shouldn't lead to any data loss.

2) "compress" - In this case, logrotate opens "log" and "log.1.gz"and runs compress program to get data for "log.1.gz". If compress program fails for some reason (including ENOSPC), we can do the same as in previous case - remove the log.1.gz and stop the rotation.

I will write some tests for that and will see. If things go well, I will commit it next week.

XufengZhang:
I'm fine with your fix, thanks a lot, hanzz!

hanzz:
I have committed my version of patch and therefore I'm closing this pull request.

PR #23, opened 2016-03-08
Remove or rename already existing file

sathieu:
Those files are created when logrotate is killed during compression or rotation.

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -367,8 +367,24 @@ static int runScript(struct logInfo *log, char *logfn, char *script)
 int createOutputFile(char *fileName, int flags, struct stat *sb, acl_type acl, int force_mode)
 {
     int fd;
-	struct stat sb_create;
-	int acl_set = 0;
+    struct stat sb_create;
+    int acl_set = 0;
+
+    if (stat(fileName, &sb_create) == 0) {
+        /* the destination file already exists, while it should not */
+        struct tm now = *localtime(&nowSecs);
+        size_t fileName_size = strlen(fileName);
+        char* backupName = alloca(fileName_size + sizeof("-YYYYMMDDHH.backup"));
+        strncpy(backupName, fileName, fileName_size);
+        size_t date_size=strftime(backupName+fileName_size, 12, "-%Y%m%d%H", &now);
+        strncpy(backupName+fileName_size+date_size, ".backup\0", 8);
+        message(MESS_ERROR, "destination %s already exists, renaming to %s\n", fileName, backupName);
+        if (rename(fileName, backupName) != 0) {
+            message(MESS_ERROR, "error renaming already existing output file %s to %s: %s\n",
+                fileName, backupName, strerror(errno));
+            return -1;
+        }
+    }
 
 	fd = open(fileName, (flags | O_EXCL | O_NOFOLLOW),
 		(S_IRUSR | S_IWUSR) & sb->st_mode);
--- a/test/test
+++ b/test/test
@@ -1814,9 +1814,36 @@ test.log.1 0 zero
 test.log.2 0 first
 EOF
 
-# check rotation with extension appended to the filename
+cleanup 72
+
+# ------------------------------- Test 72 ------------------------------------
+preptest test.log 72 2
+
+$RLR test-config.72 --force
+
+checkoutput <<EOF
+test.log 0
+test.log.1 0 zero
+test.log.2.gz 1 first
+EOF
+
+echo 'unexpected' > test.log.1.gz
+
+$RLR test-config.72 --force
+dt="$(date +%Y%m%d%H)"
+
+checkoutput <<EOF
+test.log 0
+test.log.1 0
+test.log.1.gz-$dt.backup 0 unexpected
+test.log.2.gz 1 zero
+test.log.3.gz 1 first
+EOF
+
 cleanup 100
 
+# ------------------------------- Test 100 ------------------------------------
+# check rotation with extension appended to the filename
 preptest test.log 100 1 0
 $RLR test-config.100 --force
 
--- a/test/test-config.72.in
+++ b/test/test-config.72.in
@@ -1,0 +1,7 @@
+&DIR&/test.log {
+    daily
+    rotate 3
+    compress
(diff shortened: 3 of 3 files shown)


sathieu:
Patch inspired by: https://bugs.debian.org/734688#25

glensc:
imho it's racy to `stat()` for file size and then remove it. should rename the file and then stat, but even then it is still racy if somebody has the file open. so i'd rather never remove, only rename

as it's rather bad if some process has log opened for write and the file is deleted, you have lost some "hidden" disk space which you can't reclaim back until you find and kill the offending process

sathieu:
I abandon this PR. I currently run logrotate at a different time than the reboot.

logrotate.c
```diff
@@ -363,8 +363,33 @@ static int runScript(struct logInfo *log, char *logfn, char *script)
...
+        if (sb_create.st_size == 0) {
+            /* the file is of zero size, so not needed */
+            if (unlink(fileName) != 0) {
+                message(MESS_ERROR, "error removing already existing output file %s: %s\n",
+                    fileName, strerror(errno));
+                return -1;
+            }
+        } else {
+            struct tm now = *localtime(&nowSecs);
+            size_t fileName_size = sizeof(fileName);
```
MountainRider: Should this be `size_t fileName_size = strlen(fileName);`?

logrotate.c
```diff
@@ -363,8 +363,33 @@ static int runScript(struct logInfo *log, char *logfn, char *script)
...
+            char* backupName=alloca(fileName_size+19); /* <fileName>-YYYYMMDDHH.backup */
```
MountainRider: Wouldn't this be more clear as `char* backupName = alloca(fileName_size + sizeof("-YYYYMMDDHH.backup"));`?

sathieu:
This is worst than that: If any file in a set couldn't rotate, the remaining files are NOT rotated.

I've improved the patch and provided a test for it.

sathieu:
There is still a risk to override dated backup (currently once an hour).

kdudka:
I agree that the patch introduces a TOCTOU race.  Still it looks like an improvement over status quo.  I will merge it before the next release of logrotate unless anybody comes with a better patch...

kdudka:
On Tuesday, September 13, 2016 03:08:41 Mathieu Parent wrote:
> This is worst than that: If any file in a set couldn't rotate, the remaining
> files are NOT rotated.

Is such behavior expected?  If not, it should be fixed as a separate bug IMO.

sathieu:
I don't know. shared script says: "f the scripts exit with error, the remaining actions will not be executed for any logs". So this is coherent.

kdudka:
I would expect the ```sharedscripts``` option to only affect how **scripts** are run and how their results are handled.  In my opinion, the option should **not** affect handling of other kinds of errors.

sathieu:
I've rebased the patch and added a check to a avoid renaming when destination exists.

kdudka:
Thank you for working on this!  I have pushed some further tweaks to the ```rename-existing``` branch:
https://github.com/logrotate/logrotate/compare/rename-existing
Could you please have a look whether you are fine with merging it such?

sathieu:
@kdudka Thanks. It looks good.

TafThorne:
I am pleased to find that work on fixing this seems to already happening.  I wanted to mention that in addition to [Debian Bug report:Logs are not rotated for a month](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=734688) there is [Ubuntu Bug:Logrotate fails repeatedly with /etc/cron.daily/logrotate: error: error creating output file /var/log/... File exists run-parts: /etc/cron.daily/logrotate exited with return code 1](https://bugs.launchpad.net/ubuntu/+source/logrotate/+bug/1630516) which I opened a short while ago.  

I have at least one system currently in the problem state.  If you would like to test out a proposed fix on it please let me know.

kdudka:
On Tuesday, October 18, 2016 00:39:29 Thomas Thorne wrote:
> I have at least one system currently in the problem state.  If you would
> like to test out a proposed fix on it please let me know.

That would be really appreciated.  I have just merged this into master.  
Feedback is welcome!

Qwiz:
Why does logrotate check that destination file exists? It is compressor's work, isn't it?

sathieu:
@Qwiz 
> Why does logrotate check that destination file exists? It is compressor's work, isn't it?

it does, but fails if the destination already exists. The just-merged patch moves the file before compressing which solves the failure.

Qwiz:
I've already tested it and it works. Thanks.
Will logrotate remove that backed up files after a while, or they will stay forever?

And won't it be better to add an option that allow logrotate rewriting existing files?

sathieu:
@Qwiz: the backed up filed won't be deleted. Allowing logrotate to rewrite existing files is not intended.

@kdudka about TOCTOU race, there is this post http://yakking.branchable.com/posts/moving-files-1-copying/ (found from LWN.net), and the complete source code: https://github.com/fishface60/linux-fsops

kdudka:
On Thursday, December 01, 2016 02:28:59 Alexandr Nikolaichev wrote:
> Will logrotate remove that backed up files after a while, or they will stay
> forever?

Nope.  logrotate will not remove the backed up files.  As I understand it,
the fix in question intends to resolve a fatal situation where logrotate's 
functionality is broken.  We do not expect this to happen regularly.  If it 
does, we need to look at the cause, instead of mitigating the consequences.

kdudka:
On Thursday, December 01, 2016 11:58:12 Mathieu Parent wrote:
> @kdudka about TOCTOU race, there is this post
> http://yakking.branchable.com/posts/moving-files-1-copying/ (found from
> LWN.net), and the complete source code:
> https://github.com/fishface60/linux-fsops

I see Gentoo Linux also patches logrotate to make file creation more atomic:

https://gitweb.gentoo.org/repo/gentoo.git/tree/app-admin/logrotate/files/logrotate-3.10.0-atomic-create.patch?id=3aae9fbc

I believe that logrotate-3.11.0 at least improves the situation, so I will
close this issue now.  Feel free to open a new one if the problem persists.

PR #24, opened 2016-03-15
Destination already exists message should be output at error level

lp0:
If an error like this occurs:
  destination /var/log/apport.log-201603 already exists, skipping rotation

There's no way to tell why logrotate has exited with an error. Raise the log level of this message to ERROR so that the cause of the error is available.


sergey-v:
Looks correct.

Issue #28, opened 2016-03-30
Add a "minage" feature


sastorsl:
I would like to see a "minage" feature (vs maxage) - where you can specify the time in seconds, minutes, hours or days where any file newer that the minage will be ignored/not rotated.
Seems similar to https://bugzilla.redhat.com/show_bug.cgi?id=731371

My specific use case is that I'm managing log files in several locations, and some of them are date stamped. Using the same log rotation routine / program is important.

The date stamped files are automatically created by the logger, so a new file appears at midnight.
Since I can't specify a "minage", and logrotate is run by cron.daily at about 03:15, todays logs will be spread over two files.

Moving cron.daily could fix the actual use case, but in some systems you are not able to or cron.daily should not be moved due to other considerations.

Having the possibility to ignore a file until it is old enough is a valid feature IMHO.

rvealokin:
+1 for this feature. 

In my use case I want to rotate Apache Tomcat logs. The log file names a somewhat tricky to match as some of them __always__ contain a timestamp. I would rather use a more reliable way by checking when was the last time file was modified.

rvealokin:
I have created a patch in my fork if you want to give it a try (https://github.com/rvealokin/logrotate/commit/504885b2e025928caff6dbb241f91cbe60bb7ef8).
It adds a `minage` directive where you can specify age in days.

glensc:
seems submitted as #36

PR #36, opened 2016-05-25
#28: Add a "minage" feature

inikolaev:
This adds a `minage` directive which can be used to specify number minimum file age in days relative to current date.

It checks file modification time and only rotates file when it's old enough.

I also fixed some typos and some log message was missing `\n` at the end.
Added a bit more logging to debug mode and moved number of seconds in a day into a constant.

### DIFF
--- a/config.c
+++ b/config.c
@@ -405,6 +405,7 @@ static void copyLogInfo(struct logInfo *to, struct logInfo *from)
     to->minsize = from->minsize;
 	to->maxsize = from->maxsize;
     to->rotateCount = from->rotateCount;
+    to->rotateMinAge = from->rotateMinAge;
     to->rotateAge = from->rotateAge;
     to->logStart = from->logStart;
     if (from->pre)
@@ -628,6 +629,7 @@ int readAllConfigPaths(const char **paths)
 		.minsize = 0,
 		.maxsize = 0,
 		.rotateCount = 0,
+                .rotateMinAge = 0,
 		.rotateAge = 0,
 		.logStart = -1,
 		.pre = NULL,
@@ -858,7 +860,7 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
 #endif /* MADV_DONTFORK */
 #endif /* HAVE_MADVISE */
 
-    message(MESS_DEBUG, "reading config file %s\n", configFile);
+    message(MESS_DEBUG, "Reading config file %s\n", configFile);
 
 	start = buf;
     for (start = buf; start - buf < length; start++) {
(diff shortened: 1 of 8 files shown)


kdudka:
I have rebased the above commits on top of current master:

https://github.com/logrotate/logrotate/compare/minage

Could you please have a look?

kdudka:
@glensc Do you feel like doing a review of https://github.com/logrotate/logrotate/compare/minage?

glensc:
@kdudka looks simple and good. did not spot any obvious mistakes. but didn't look closely the config parsing part. i assume it's just copy paste of similar parsing of other directives.

aside i noticed, #28 mentions about using seconds, but this PR (and your branch) min granularity is actually full days (can't even tweak to use `minage 0.5` to specify 12 hours :D).

anyway, LGTM for merge!

kdudka:
On Thursday, September 08, 2016 12:36:49 Elan Ruusam�e wrote:
> @kdudka looks simple and good. did not spot any obvious mistakes. but didn't
> look closely the config parsing part. i assume it's just copy paste of
> similar parsing of other directives.

That was fast.  Thanks for review!

> aside i noticed, #28 mentions about using seconds, but this PR (and your
> branch) min granularity is actually full days (can't even tweak to use
> `minage 0.5` to specify 12 hours :D).

True.  But we have the same problem with the "maxage" directive already.

> anyway, LGTM for merge!

Perfect.  I will merge it then...

tomaszpr:
Hi, in my opinion it doesn't work as expect.
According to code:

```
if (log->rotateMinAge && log->rotateMinAge >= (nowSecs - sb.st_mtime) / DAY_SECONDS) {
	    state->doRotate = 0;
	    message(MESS_DEBUG, "  log does not need rotating "
		    "('minage' directive is used and the log "
		    "age is smaller than the minage days)\n"); 

```
When I set minage 1, file will be rotated after 2 days, not after 1 day.

Example:
**rotateMinAge = 1**
Last file change: 12:00:00 
Current time:  12:01:00 next day
Time diff: 86400+60 > 1 day , in my opinion rotation should be done, but:
(nowSecs - sb.st_mtime) = 86400+60
(nowSecs - sb.st_mtime) / DAY_SECONDS = 1

if (rotateMinAge >= 1) !!!
state->doRotate = 0; DO NOT ROTATE!!!!

Rotation will be done after two all days, but I set minage 1.

inikolaev:
I think you are right - it doesn't calculate age properly. I guess a better way would be:

```
log->rotateMinAge * DAY_SECONDS >= nowSecs - sb.st_mtime
```
in this case it would not be affected by this rounding issue.

kdudka:
It makes sense to me.  @inikolaev should I push the change on your behalf?

inikolaev:
Yes, go ahead @kdudka

PR #37, opened 2016-05-28
Fix trivial typo in debug message.

goeranu:
Not much to add to the checkin message.  I just happened to see it when looking for a problem in my configuration.


rvealokin:
I think this has already been addressed in several pull requests :)

kdudka:
Sorry for replying late!  This is now fixed via fe80cb51a2571ca35b1a7c8ba0695db5a68feaba and it will be included in the upcoming release.

PR #40, opened 2016-06-11
Fix maxage calculation

### LINKED ISSUE #39: maxage calculation is off-by-1
With maxage set, any log file older than N days needs to be purged. In effect, any file N+1 days old and older is purged (for example, with maxage set to 1, a file 25 hours old is not deleted - it is only deleted if it's 48 hours old or older).

urisimchoni:
This patch fixes maxage calculation (fixes #39), and adds some tests for maxage.

Thanks,
Uri.


urisimchoni:
Closing the request since the relevant issue is closed - logrotate behaves as designed.

PR #46, opened 2016-08-02
travis: make distcheck

glensc:
run make dist on travis

i wanted to add `make distcheck`, but `make test` is already ran in previous step

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ script:
   - make
   - make test
   - cd test && /bin/dash test && cd ..
+  - make distcheck
 branches:
   only:
     - master


kdudka:
The problem is that ```make dist``` succeeds on the ```3.9.2``` release of logrotate, too.  The fact that ```make dist``` succeeds does not imply that the resulting tarball is anyhow usable.  In contrast to that, ```make distcheck``` checks much more.  For example, it tests that out-of-source build succeeds if the source directory tree is read-only.  We should really include ```make distcheck``` in CI once we get it running in the CI environment.

glensc:
replaced with make distcheck!

PR #49, opened 2016-08-02
Drop root Makefile

glensc:
Makefile is generated by autoconf/automake


kdudka:
Good point.  The legacy ```Makefile``` is not packaged into the release tarball produced by ```make dist```.  However, I am not sure whether we want to drop it altogether.  It still appears to work and people who do not like autotools might be still using it.  Do you really think it is safe to drop it now?

glensc:
why keep it if logrotate maintainers to do support it. also it may be mistakenly committed over with autofoo version. end users should use release tarball (not git master) which contains automake/autoconf version of the makefile.

glensc:
and it's super annoying if you are really developing logrotate. i have to git stash or git reset every time i wish to pull in new changes because the changes in Makefile prevent from git pull succeeding. and then updating upstream git changes need to run autogen again because restored makefile is not from automake...: http://sprunge.us/BFjK

kdudka:
I agree that, if a git-tracked file is overwritten by generated file while running ```./configure```, it causes a lot of problems to logrotate developers.  The question is how to fix it:

1. We can just stop maintaining/distributing the legacy Makefile as you proposed originally.
2. But we can also rename/move it elsewhere and keep it around for people who prefer using readable Makefiles.

I have no strong preference on that, just wanted to point out that removing the legacy Makefile is not the only correct solution.

glensc:
i'm ok with renaming it to `Makefile.legacy` and appropriate doc updates indicate to run `make -f Makefile.legacy`

kdudka:
OK.  I will rename it to ```Makefile.legacy``` and include it in the release tarball.  As the Makefile is deprecated, there is no documentation to update.

PR #35, opened 2016-05-19
fixed compile warning with gcc6

cromerc:
Now in gcc6 it gives warnings on compile when this type of code is used:
if (something == 0) 
  return 0;

The fix is simple:
if (something == 0) {
  return 0;
}

PR #50, opened 2016-08-02
fixed gcc6 warning

cromerc:
Here is the patch to fix the warning under gcc6.


kdudka:
Thanks for resubmitting it!  I will merge the patch as is.  Next time please include the related part of compiler's error output into the commit message.

glensc:
to get previous PR linked, this is re-submit of #35

@cromerc next time include such info in description :)

PR #51, opened 2016-08-03
largefile support configure test

glensc:
some old patch from my distro:
https://github.com/pld-linux/logrotate/commits/auto/th/logrotate-3.9.2-1/logrotate-largefile.patch

as it is autoconf test, should be very portable and safe

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -10,6 +10,8 @@ AC_PROG_CC_STDC
 AC_STRUCT_ST_BLKSIZE
 AC_STRUCT_ST_BLOCKS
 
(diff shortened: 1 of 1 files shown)


kdudka:
Thanks for the patch!  What do we need it for?  How can we verify that it works actually?

glensc:
try to create file bigger than 2GiB on 32bit system and let logrotate handle it. i assume size checks will fail, likely other operations (rotation based on timestamp) will still succeed. it's generally good idea to have large file support. the patch has been applied for almost two years and have not heard complaints.

PR #52, opened 2016-08-04
Fix the download link in README.md

ahpnils:
The link for logrotate-3.10.0 is actually for 3.9.2.
The good news is, the changelog is alright :)


kdudka:
Embarrassing.  Thanks for the fixup!

PR #53, opened 2016-08-12
BUG: support for usernames starting with numeric symbols

EwaldvanGeffen:


kdudka:
Thanks for the patch!  I have just moved the check a little bit up to make it clear that the first ```sscanf``` is not needed for the ```su``` directive.

PR #54, opened 2016-08-16
test: fix reported test number when skipping test 46

fishilico:
When running "make test" on a system where SELinux is disabled, the
output is currently:

    ...
    Running test 45
    Running test 46
    Skipping SELinux part of test 6
    Running test 47
    Skipping SELinux test 47
    Running test 48
    ...

The test number in the message should be 46 instead of 6.


kdudka:
Indeed.  Thanks for the patch!

PR #55, opened 2016-08-19
Use autoconf checks for strndup and asprintf

tpetazzoni:
The current code in config.c can provide its own implementation of
asprintf() and strndup() if not provided by the system. However, in
order to decide if they should be provided, the check done is:

 #if !defined(name_of_function)

which only works if the function is actually defined as a macro, which
is not necessarily the case.

Therefore, we replace this logic by a proper AC_CHECK_FUNCS() check in
the configure script.

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@free-electrons.com>

### DIFF
--- a/config.c
+++ b/config.c
@@ -45,7 +45,7 @@
 #include "asprintf.c"
 #endif
 
-#if !defined(asprintf) && !defined(_FORTIFY_SOURCE)
+#if !defined(HAVE_ASPRINTF) && !defined(_FORTIFY_SOURCE)
 #include <stdarg.h>
 
 int asprintf(char **string_ptr, const char *format, ...)
@@ -78,7 +78,7 @@ int asprintf(char **string_ptr, const char *format, ...)
 
 #endif
 
-#if !defined(strndup)
+#if !defined(HAVE_STRNDUP)
 char *strndup(const char *s, size_t n)
 {
        size_t nAvail;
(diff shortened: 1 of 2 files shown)


kdudka:
Thanks for the patch!  Just out of curiosity, what is the system where you need local implementation of those functions?

tpetazzoni:
@kdudka What do you mean by "local" implementation? If you mean the implementation of strndup and asprintf provided as part of the logrotate code, I don't need them at all: they are already provided by the three standard C libraries commonly used on Linux systems (glibc, uClibc, musl).

My patch precisely aims at using the functions provided by the C library, rather than the ones provided by logrotate.

kdudka:
Indeed.  While trying it locally, I used ```glibc``` and ```gcc -O2```, which caused ```strndup``` to be defined as a macro.  Thank you for clarifying it!

tpetazzoni:
Might be the case with glibc, but with musl and uClibc, they are not necessarily defined as macros. In any case, using an autoconf test is a lot more portable than doing a preprocessor test on the function name.

kdudka:
On Monday, August 22, 2016 10:26:48 Thomas Petazzoni wrote:
> In any case, using an autoconf test is a lot more portable
> than doing a preprocessor test on the function name.

Definitely.  Thanks for the contribution!

PR #59, opened 2016-09-19
BUG/MAJOR: config.c: re-stat() path after mkdir()

bjacquin:
If a directory is created (ie createolddir), struct sb must be updated
in order to get appropriate st_uid and st_gid.

Before:
  ./logrotate  -f /etc/logrotate.conf
  uid: 250, sb.st_uid: 250
  gid: 250, sb.st_gid: 250

After:
  ./logrotate  -f /etc/logrotate.conf
  uid: 250, sb.st_uid: 0
  gid: 250, sb.st_gid: 250

### DIFF
--- a/config.c
+++ b/config.c
@@ -318,8 +318,7 @@ static int do_mkdir(const char *path, mode_t mode, uid_t uid, gid_t gid) {
 				path, strerror(errno));
 			return -1;
 		}
-		if ((uid != sb.st_uid || gid != sb.st_gid) && 
-			chown(path, uid, gid)) {
+		if (chown(path, uid, gid)) {
 			message(MESS_ERROR, "error setting owner of %s to uid %d and gid %d: %s\n",
 				path, uid, gid, strerror(errno));
 			return -1;


glensc:
you must include config which the problem can be reproduced

kdudka:
Travis CI revealed that the newly added code does not compile because a semicolon is missing:
https://travis-ci.org/logrotate/logrotate/jobs/160922981

kdudka:
As @bjacquin pointed out, the code is obviously wrong.  The condition compares the requested uid/gid with ```stat``` buffer containing undefined values because ```stat()``` has failed.  I believe that author's intention was to compare uid/gid in case the file already exists and adjust accordingly, but:
1) This has never been documented, neither actually worked.
2) It introduces yet another TOCTOU race.

So I would just  remove the bogus condition ```(uid != sb.st_uid || gid != sb.st_gid)``` to always call ```chown()``` after ```mkdir()```.
@glensc Do you agree?

kdudka:
@glensc The configuration directive in question is ```createolddir```.

glensc:
should the method do `chown` even in case if directory exists? or only when directory is created?

i don't know which is the uid if that code block is ran, but if you always `chown` you may get permission errors which you did not get earlier because `chown` was ran only on specific conditions.

kdudka:
On Monday, September 19, 2016 08:09:03 Elan Ruusam�e wrote:
> should the method do `chown` even in case if directory exists?

As far as I understand, the documentation does not imply anything like that.

> or only when directory is created?

Yes, that is how ```createolddir``` is currently documented.

> i don't know which is the uid if that code block is ran, but if you always
> `chown` you may get permission errors which you did not get earlier because
> `chown` was ran only on specific conditions.

I believe the condition was met in most of the cases.  In other words, it is 
unlikely that the random stack contents would match the given uid/gid and the 
```chown()``` was skipped.

glensc:
i'm ok then with your suggestion to drop comparing undefined stat buffer and always `chown(2)`

bjacquin:
Indeed, there was a comma missing after a last minute rebase, my apologies. Is my config still needed ?

I updated this PR to remove tests and always perform chown()

PR #64, opened 2016-09-23
convert changelog to markdown

glensc:
tries to follow to [Keep a CHANGELOG](http://keepachangelog.com/) guidelines.

should be squashed when merging, but you squash them anyway when merging to master. so the separate commits can be kept around to see the actual changes

took the release dates from git commit date (not author date).

there also may be difference in the historic releases when actual upload happened latter. but i guess it's not that important anymore the exact date.


kdudka:
Not sure I understand.  I prefer separate commits as long as the project is consistent at each point in the history (in other words, git-bisect ready).  If there is a fixup commit (e.g. the one fixing the build system), I prefer to squash it before merge.

glensc:
i mean you don't do merge commits, as i assume you like flat project history.

anyway, the changes here are atomic enough. do you want me to squash any of the commits?

glensc:
as for bisect ready, i guess i should squash these two:

* 0cdc2e7 - include ChangeLog.md in dist (41 minutes ago) <Elan Ruusamäe>
* 46bddd5 - rename changelog to .md (2 hours ago) <Elan Ruusamäe>

kdudka:
Exactly.  I will take care of the squashing.  Thank you for the contribution!

PR #65, opened 2016-09-23
Create CONTRIBUTING.md

glensc:
adds CONTRIBUTING.md: https://github.com/blog/1184-contributing-guidelines

as per [discussion](https://github.com/logrotate/logrotate/commit/b4c9091fc011ec9c449553ea8bfc5a2b04d887ce#commitcomment-19150719)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,7 +5,12 @@
   - Fork it.
   - Create your feature branch (`git checkout -b fixing-blah`), please avoid working directly on the `master` branch.
   - Check for unnecessary whitespace with `git diff --check` before committing.
-  - Add your change to `ChangeLog.md` under the `UNRELEASED` section
-  - Commit your changes (`git commit -am 'Fixed blah'`).
+  - [optinal] Add user visible changes to `ChangeLog.md` under the `UNRELEASED` section
+  - Commit your changes, try to follow this format: 
+```
+scope: short summary of the change
+[empty line]
+Long description of the change, explanation of why the change is useful, etc.
+```
   - Push to the branch (`git push -u origin fixing-blah`).
   - Create a new pull request.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,11 @@
+# logrotate contributions
+
+## Pull requests
+
+  - Fork it.
+  - Create your feature branch (`git checkout -b fixing-blah`), please avoid working directly on the `master` branch.
+  - Check for unnecessary whitespace with `git diff --check` before committing.
+  - Add your change to `ChangeLog.md` under the `UNRELEASED` section
+  - Commit your changes (`git commit -am 'Fixed blah'`).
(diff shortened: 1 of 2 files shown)


glensc:
here's little chicken-egg problem:
1. user should write changelog and would be nice to include PR number
2. user has not submitted PR yet, so no PR number present

so this could be solved as:
1. create PR
2. add changelog entry, amend last commit, push -f

but that is way too complicated to describe and average user to follow

perhaps:
1. create PR
2. add changelog entry with PR number and credit yourself, example
```
## UNRELEASED

  - convert changelog to markdown. (#64, @glensc)
```

but that creates additional commit, is that ok?

glensc:
perhaps also add something:

```
  - add more commits to your PR, push to same branch
```

kdudka:
On Friday, September 23, 2016 13:23:05 Elan Ruusam�e wrote:
> here's little chicken-egg problem:
> 1. user should write changelog and would be nice to include PR number
> 2. user has not submitted PR yet, so no PR number present
> 
> so this could be solved as:
> 1. create PR
> 2. add changelog entry, amend last commit, push -f
> 
> but that is way too complicated to describe and average user to follow

Yes it is.  This can be trivially solved by the person who pushes the commits 
upstream, just before they are pushed.  Moreover, it is common that one commit 
closes multiple issues or pull requests, where some of them an occasional 
contributor does not even know about.

> perhaps:
> 1. create PR
> 2. add changelog entry with PR number and credit yourself, example
> ```
> ## UNRELEASED
> 
>   - convert changelog to markdown. (#64, @glensc)
> ```
> 
> but that creates additional commit, is that ok?

It would not solve the problem because we want to reference the discussion, 
review comments etc. from the actual commit that _resolved_ an issue, not a 
random commit around it.

CONTRIBUTING.md
```diff
@@ -0,0 +1,11 @@
+# logrotate contributions
+
+## Pull requests
+
+  - Fork it.
+  - Create your feature branch (`git checkout -b fixing-blah`), please avoid working directly on the `master` branch.
+  - Check for unnecessary whitespace with `git diff --check` before committing.
+  - Add your change to `ChangeLog.md` under the `UNRELEASED` section
```
kdudka: I would prefer to add only **user visible changes** to ChangeLog.md.  Implementation details can be easily found in git history.

glensc:
well. let's leave it as is then.

you can remove `- Add your change to ChangeLog.md under the UNRELEASED section` from the file if it that brings more confusion than gain

CONTRIBUTING.md
```diff
@@ -0,0 +1,11 @@
...
+  - Commit your changes (`git commit -am 'Fixed blah'`).
```
kdudka: Should we more precisely specify the commit message format?  Something like:
```
scope: short summary of the change
[empty line]
Long description of the change, explanation of why the change is useful, etc.
```

glensc: updated

PR #66, opened 2016-10-03
Fix test suite in combination with bash 4.4

DimStar77:
Until bash 4.4, variables weresilently striped of NUL bytes; with Bash 4.4
this results now for every single 0-byte that is being stripped in a warning:

./test: line 170: warning: command substitution: ignored null byte in input

We now simply strip the NUL bytes before we have bash assign the values to
a variable. The result is the same.

### DIFF
--- a/test/test
+++ b/test/test
@@ -183,7 +183,7 @@ checkoutput() {
 	if [ -n "$co_compressed" ] && [ "$co_compressed" != 0 ]; then
 		contents=`gunzip -c $file`
 	else
-		contents=`cat $file`
+		contents=`cat $file | tr -d '\000'`
 	fi
 	if [ "$contents" != "$expected" ]; then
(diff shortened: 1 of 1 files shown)


glensc:
you should avoid working on master branch, see [CONTRIBUTING.md](https://github.com/logrotate/logrotate/blob/master/CONTRIBUTING.md)

the problem is that all commits associated to the branch this PR was created from get attached to the PR. meaning your `master` branch is locked out until the PR is merged, or you will be pushing irrelevant commits to specific PR.

but this PR is probably simple and maintainer can merge it, and you can live with this one time mistake

DimStar77:
I wasn't planning on immediate more fixes on logrotate - so that should be safe

kdudka:
Looks good to me.  Thanks!

PR #68, opened 2016-10-05
Patches that are in use within openSUSE

scarabeusiv:
While updating to latest logrotate I noticed that in openSUSE we carryover quite few patches and thus took it on me to push them forward :)

We also have for last 10 years patch to determine compression extension, but I am not sure how welcome that would be as it is not acompanied by uinttest:
https://build.opensuse.org/package/view_file/Base:System/logrotate/logrotate-autoext.patch?expand=1

### DIFF
--- a/examples/logrotate.cron
+++ b/examples/logrotate.cron
@@ -1,8 +1,23 @@
-#!/bin/sh
+#!/bin/bash
 
-/usr/sbin/logrotate /etc/logrotate.conf
-EXITVALUE=$?
-if [ $EXITVALUE != 0 ]; then
-    /usr/bin/logger -t logrotate "ALERT exited abnormally with [$EXITVALUE]"
+# exit immediately if there is another instance running
+if pgrep /usr/sbin/logrotate > /dev/null ; then
+	/bin/logger -p cron.warning -t logrotate "ALERT another instance of logrotate is running - exiting"
+	exit 1
 fi
+
+TMPF=`mktemp ${TMPDIR:-/tmp}/logrotate.XXXXXXXXXX`
+
+/usr/sbin/logrotate /etc/logrotate.conf 2>&1 | tee $TMPF
+EXITVALUE=${PIPESTATUS[0]}
+
+if [ $EXITVALUE != 0 ]; then
+    # wait a sec, we might just have restarted syslog
+    sleep 1
+    # tell what went wrong
+    /bin/logger -p cron.warning -t logrotate "ALERT exited abnormally with [$EXITVALUE]"
+    /bin/logger -p cron.warning -t logrotate -f $TMPF
+ fi
+
+rm -f $TMPF
 exit 0
--- a/config.c
+++ b/config.c
@@ -648,6 +648,7 @@ int readAllConfigPaths(const char **paths)
 		.preremove = NULL,
 		.logAddress = NULL,
 		.extension = NULL,
+		.addextension = NULL,
 		.compress_prog = NULL,
 		.uncompress_prog = NULL,
 		.compress_ext = NULL,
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
+*.orig
+*.rej
```
kdudka: Those files are not generated by logrotate's build system.  Hence, in my opinion, they do not belong to logrotate's ```.gitignore```.  But the same holds for ```.*.swp``` or ```cscope.out```, which are already there.

@glensc, what is your opinion on this?

kdudka:
On Wednesday, October 05, 2016 05:09:21 Tomáš Chvátal wrote:
> We also have for last 10 years patch to determine compression extension, but
> I am not sure how welcome that would be as it is not acompanied by
> uinttest:
> https://build.opensuse.org/package/view_file/Base:System/logrotate/logrotate-autoext.patch?expand=1

Please submit it as a separate pull request.  The fprintf(stdout, ...) calls
are there intentionally or is it just leftover from some debugging?

.gitignore
glensc: i would allow it, altho preferred way is that people setup their editor backup files globally ([SO post](http://stackoverflow.com/questions/7335420/global-git-ignore))

examples/logrotate.cron
```diff
@@ -1,8 +1,23 @@
...
+# exit immediately if there is another instance running
+if checkproc /usr/sbin/logrotate; then
+	/bin/logger -p cron.warning -t logrotate "ALERT another instance of logrotate is running - exiting"
+	exit 1
 fi
+
+TMPF=`mktemp /tmp/logrotate.XXXXXXXXXX`
+
+/usr/sbin/logrotate /etc/logrotate.conf 2>&1 | tee $TMPF
+EXITVALUE=${PIPESTATUS[0]}
```
glensc: `$PIPESTATUS[0]` is considered bashism, and the `#!` still points to plain `sh`

glensc:
even i commented various patches, i think this PR should add only one feature and the other features in separate PR. this one could be feature adding addextension.

examples/logrotate.cron
```diff
@@ -1,8 +1,23 @@
 #!/bin/sh
 
-/usr/sbin/logrotate /etc/logrotate.conf
-EXITVALUE=$?
-if [ $EXITVALUE != 0 ]; then
-    /usr/bin/logger -t logrotate "ALERT exited abnormally with [$EXITVALUE]"
+# exit immediately if there is another instance running
+if checkproc /usr/sbin/logrotate; then
```
glensc: the whole example script seems distribution specific. perhaps another file if it's accepted to include it in upstream

.gitignore
kdudka: Then I prefer not to apply the patch.  It adds no value to logrotate.  Moreover, the presence of ```*.orig``` and ```*.rej``` usually indicates that some patches did not apply properly.  Hiding it can be dangerous in certain situations.

kdudka:
On Wednesday, October 05, 2016 09:35:50 Elan Ruusam�e wrote:
> even i commented various patches, i think this PR should add only one
> feature and the other features in separate PR. this one could be feature
> adding addextension.

As long as we can work with the patches separately, I am fine with PRs
like this.  Creating one pull-request per issue could overly bother our 
contributors.

All the pull requests are github-specific and we have no guarantee that
github will exist forever, or that the data from our pull requests could
ever be migrated to something else.  So I would not insist on spending
too much effort on github PRs.  I believe it is the git repository itself
what matters from the long term perspective.

scarabeusiv:
Rebased including the comments from above.

examples/logrotate.cron
```diff
@@ -1,8 +1,23 @@
...
 
-/usr/sbin/logrotate /etc/logrotate.conf
-EXITVALUE=$?
-if [ $EXITVALUE != 0 ]; then
-    /usr/bin/logger -t logrotate "ALERT exited abnormally with [$EXITVALUE]"
+# exit immediately if there is another instance running
+if pgrep /usr/sbin/logrotate > /dev/null ; then
+	/bin/logger -p cron.warning -t logrotate "ALERT another instance of logrotate is running - exiting"
+	exit 1
 fi
```
glensc: also, you changed path to `logger` binary in this commit. perhaps just rely on `$PATH` instead of hardcoding tool paths?

kdudka: In my opinion, it would be safer to store logrotate's PID to a file and compare both PID and process name on the subsequent run (in case the PID file still exists).  Not sure if we can portably implement this without extending logrotate to optionally create the PID file...

PR #69, opened 2016-10-06
Compression autoext

scarabeusiv:
This determines suffix for files that undergo compression.
Depending on the compressor you get the correct suffix.

### DIFF
--- a/config.c
+++ b/config.c
@@ -727,6 +727,7 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
     char **scriptDest = NULL;
     struct logInfo *newlog = defConfig;
     char *start, *chptr;
+    char *compresscmd_base;
     char *dirName;
     struct passwd *pw = NULL;
     int rc;
@@ -1266,6 +1267,20 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
 					message(MESS_DEBUG, "compress_prog is now %s\n",
 						newlog->compress_prog);
 
+			 		compresscmd_base=strdup(basename(newlog->compress_prog));
+			 		fprintf(stdout, "compress_ext is %s\n", newlog->compress_prog);
+			 		i=0; /* have to check whether we may do this! */
+			 		/* we check whether we changed the compress_cmd. In case we use the apropriate extension
+			 		   as listed in compress_cmd_list */
+			 		while ((i>=0)&&(strcmp(compress_cmd_list[i][0], "EOLIST"))){
+			 		    if (0==strcmp(compress_cmd_list[i][0], compresscmd_base)){
+			 			newlog->compress_ext=strdup((char *)compress_cmd_list[i][1]);
+			 			message(MESS_DEBUG, "compress_ext was changed to %s\n", newlog->compress_ext);
+			 			fprintf(stdout, "compress_ext was changed to %s\n", newlog->compress_ext);
+			 			i=-10; /* terminate loop! */
+			 		    }
+			 		    i++;
+			 		}
 				} else if (!strcmp(key, "uncompresscmd")) {
 					freeLogItem (uncompress_prog);
 
--- a/config.c
+++ b/config.c
@@ -1279,7 +1279,7 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
 			 		compresscmd_base=strdup(basename(newlog->compress_prog));
 			 		/* we check whether we changed the compress_cmd. In case we use the apropriate extension
 			 		   as listed in compress_cmd_list */
-					for(i=0; compress_cmd_list[i][0] == NULL; i++) {
+					for(i=0; compress_cmd_list[i][0] != NULL; i++) {
 						if (!strcmp(compress_cmd_list[i][0], compresscmd_base)) {
 							newlog->compress_ext=strdup((char *)compress_cmd_list[i][1]);
 							message(MESS_DEBUG, "compress_ext was changed to %s\n", newlog->compress_ext);
--- a/config.c
+++ b/config.c
@@ -1268,7 +1268,6 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
 						newlog->compress_prog);
 
 			 		compresscmd_base=strdup(basename(newlog->compress_prog));
-			 		fprintf(stdout, "compress_ext is %s\n", newlog->compress_prog);
 			 		i=0; /* have to check whether we may do this! */
 			 		/* we check whether we changed the compress_cmd. In case we use the apropriate extension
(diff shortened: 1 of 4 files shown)


config.c
```diff
@@ -727,6 +727,7 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
     char **scriptDest = NULL;
     struct logInfo *newlog = defConfig;
     char *start, *chptr;
+    char *compresscmd_base;
```
kdudka: Please move the variable definition to the innermost code block to limit its scope.

config.c
```diff
@@ -1266,6 +1267,20 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
 					message(MESS_DEBUG, "compress_prog is now %s\n",
 						newlog->compress_prog);
 
+			 		compresscmd_base=strdup(basename(newlog->compress_prog));
+			 		fprintf(stdout, "compress_ext is %s\n", newlog->compress_prog);
+			 		i=0; /* have to check whether we may do this! */
```
kdudka: Why assigning zero to the variable ```i``` would be a problem?  Is the comment misplaced?

config.c
```diff
@@ -1266,6 +1267,20 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
...
+			 		/* we check whether we changed the compress_cmd. In case we use the apropriate extension
+			 		   as listed in compress_cmd_list */
+			 		while ((i>=0)&&(strcmp(compress_cmd_list[i][0], "EOLIST"))){
+			 		    if (0==strcmp(compress_cmd_list[i][0], compresscmd_base)){
+			 			newlog->compress_ext=strdup((char *)compress_cmd_list[i][1]);
+			 			message(MESS_DEBUG, "compress_ext was changed to %s\n", newlog->compress_ext);
+			 			fprintf(stdout, "compress_ext was changed to %s\n", newlog->compress_ext);
+			 			i=-10; /* terminate loop! */
```
kdudka: Please use ```break``` to terminate the loop.

config.c
```diff
@@ -1266,6 +1267,20 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
...
+			 		    }
+			 		    i++;
```
kdudka: Please use the ```for``` statement for a counted loop.

logrotate.c
```diff
@@ -106,6 +106,16 @@ unsigned int hashSize;
...
 int debug = 0;
 char *mailCommand = DEFAULT_MAIL_COMMAND;
+/* This is the declaration. Note the "2" is needed in the definition (logrotate.h), too! */
+const char * compress_cmd_list[][2] = {
+  {"gzip", ".gz"},
+  {"bzip2", ".bz2"},
+  {"xz", ".xz"},
+  {"compress", ".Z"},
+  {"zip", "zip"},
+  {"EOLIST", "EOLIST"} /* end-marker */
```
kdudka: Would not it be both more efficient and more readable to use {NULL, NULL} as the end-marker?

logrotate.c
```diff
@@ -106,6 +106,16 @@ unsigned int hashSize;
 int numLogs = 0;
 int debug = 0;
 char *mailCommand = DEFAULT_MAIL_COMMAND;
+/* This is the declaration. Note the "2" is needed in the definition (logrotate.h), too! */
+const char * compress_cmd_list[][2] = {
```
kdudka: Why do you define the array in a different module than the one that actually uses it?  Please make it a static array in the appropriate module to limit visibility of the symbol.

logrotate.h
```diff
@@ -80,6 +80,9 @@ TAILQ_HEAD(logInfoHead, logInfo) logs;
 extern int numLogs;
 extern int debug;
 
+/* This is the definition. Note we have to carry the "2" within the declaration (logrotate.c), too! */
+extern const char * compress_cmd_list[][2];
```
kdudka: Please drop this.  We do not want locally used symbols to be globally available.

test/test
```diff
@@ -577,32 +577,6 @@ checkoutput <<EOF
 test.log 0 zero
 EOF
 
-cleanup 18
```
kdudka: You are removing an existing test-case without any comment.  Are you going to break the scenario covered by the test-case without telling potential users about the breakage?

config.c
```diff
@@ -1266,6 +1267,20 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
 					message(MESS_DEBUG, "compress_prog is now %s\n",
 						newlog->compress_prog);
 
+			 		compresscmd_base=strdup(basename(newlog->compress_prog));
+			 		fprintf(stdout, "compress_ext is %s\n", newlog->compress_prog);
+			 		i=0; /* have to check whether we may do this! */
+			 		/* we check whether we changed the compress_cmd. In case we use the apropriate extension
+			 		   as listed in compress_cmd_list */
+			 		while ((i>=0)&&(strcmp(compress_cmd_list[i][0], "EOLIST"))){
```
kdudka: Please delimit binary operators by spaces as done in the surrounding code.

config.c
```diff
@@ -1268,16 +1277,14 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
...
 			 		/* we check whether we changed the compress_cmd. In case we use the apropriate extension
 			 		   as listed in compress_cmd_list */
-			 		while ((i>=0)&&(strcmp(compress_cmd_list[i][0], "EOLIST"))){
-			 		    if (0==strcmp(compress_cmd_list[i][0], compresscmd_base)){
-			 			newlog->compress_ext=strdup((char *)compress_cmd_list[i][1]);
-			 			message(MESS_DEBUG, "compress_ext was changed to %s\n", newlog->compress_ext);
-			 			i=-10; /* terminate loop! */
-			 		    }
-			 		    i++;
+					for(i=0; compress_cmd_list[i][0] == NULL; i++) {
```
kdudka: The loop condition seems to be negated by mistake.  How are you confirming that the resulting code works as expected after your recent changes?

scarabeusiv:
Good point about it being inverted. I used the test after the changes to see if it is still failing, but then it worked and since I had bit of time when I didn't bother to think about it my thinking was more in a way of 'yay it works'.
Okay now it should really do what is expected.

kdudka:
Thanks for the update!  I have prepared this PR for merge in the ```autoext``` branch:
https://github.com/logrotate/logrotate/compare/autoext
Are you fine with merging it such?

scarabeusiv:
Sure, it does the thing what the patch set about -> go ahead and finaly no left patches at SUSE side :)

PR #70, opened 2016-10-06
buildsystem: use serial-harness automake feature

praiskup:


glensc:
this seems to be deprecated feature:
https://www.gnu.org/software/automake/manual/html_node/Serial-Test-Harness.html#Serial-Test-Harness

yet the documentation itself states it's easier to implement that `Parallel Tests`

praiskup:
Yep, you should use parallel-harness if possible, but AFAIK it doesn't mean that serial-harness goes away from automake -- perhaps that's just a bad wording in automake manual.  But there is also written that there may exist a good reasons to use serial-harness now.

The point is that parallel-harness drops stdout (that's expected because multiple sub-processes writing into stdout would cause ugly log output), serial harness keeps showing standard output.

praiskup:
@glensc, this patch is enough to switch into "parallel-harness" btw:
```patch
-AM_INIT_AUTOMAKE([foreign serial-tests])
+AM_INIT_AUTOMAKE([foreign])
```
But to benefit from this, somebody would need to split the testsuite properly.  Looking at the test directory, it looks like it could be done...  yes, I could help with Automake details, but not with the tests, as I don't know logrotate at all.

PR #76, opened 2016-10-19
Gitignore

dougsland:


dougsland:
The adjust of Makefile in gitignore is not correct because it will affect the 'make test'

PR #79, opened 2016-10-21
build: use autotools instead of config.h

dougsland:
With autotools we can determine an unique source of data
to fill .c, .h, .spec or even man pages. Previously,
logrotate had ifdef's to hold data but because it's not dynamic
some places of source code had to duplicate it statically.


dougsland:
Travis issue not related to the patch:

<snip>
$ git clone --depth=50 https://github.com/logrotate/logrotate.git logrotate/logrotate

Cloning into 'logrotate/logrotate'...

fatal: unable to access 'https://github.com/logrotate/logrotate.git/': Couldn't resolve host 'github.com'

The command "eval git clone --depth=50 https://github.com/logrotate/logrotate.git logrotate/logrotate" failed. Retrying, 2 of 3.

Cloning into 'logrotate/logrotate'...

fatal: unable to access 'https://github.com/logrotate/logrotate.git/': Couldn't resolve host 'github.com'

The command "eval git clone --depth=50 https://github.com/logrotate/logrotate.git logrotate/logrotate" failed. Retrying, 3 of 3.

Cloning into 'logrotate/logrotate'...

fatal: unable to access 'https://github.com/logrotate/logrotate.git/': Couldn't resolve host 'github.com'

The command "eval git clone --depth=50 https://github.com/logrotate/logrotate.git logrotate/logrotate" failed 3 times.

The command "git clone --depth=50 https://github.com/logrotate/logrotate.git logrotate/logrotate" failed and exited with 128 during .

Your build has been stopped.
</snip>

glensc:
i wouldn't rename the `.c` files, instead use defines from `config.h.in`. it makes coding really difficult.

kdudka:
On Sunday, October 23, 2016 11:26:00 Elan Ruusam�e wrote:
> i wouldn't rename the `.c` files, instead use defines from `config.h.in`.

Exactly.  It is config.h what is usually generated by the configure script.  
Substitutions in man pages are fine but having C units generated feels wrong 
to me and would introduce a lot of troubles.

dougsland:
Sure, just tried to simplify the source. I will prepare a new commit soon this week.

kdudka:
Let me close this ticket as we agreed that it does not take the correct approach.  I will prepare a patch that only makes the state file path configurable...

PR #91, opened 2016-12-03
changelog: update 3.11.0 links

glensc:
this should had been done before tagging 3.11.0 so that link would appear in released tarball.


glensc:
Travis job failure is caused by temporary Travis network issue.

PR #94, opened 2016-12-06
GPG instructions

g4jc:
Added GPG information to README.md.


glensc:
@g4jc here's some logic error: you shouldn't even unpack the tarball if the signature verification fails :)

g4jc:
@glensc My apologies, I was basing my text on a tutorial and it was wrong! I will open a new pull request with fixed instructions.

PR #96, opened 2016-12-07
improve gpg instructions

kerolasa:
Avoid hash collision vulnerability, and tell how to check git tag signature in README file.


glensc:
thanks for the daily readme update! :D

PR #98, opened 2016-12-15
Fix warnings

kerolasa:
Makes the following to pass without errors, or breaking tests.

CFLAGS='-O -g -ggdb -Wall -Wextra -fno-omit-frame-pointer -Wshadow -D_FORTIFY_SOURCE=2 -fstack-protector-all -Wno-deprecated -Wno-deprecated-declarations -Wnull-dereference'

### DIFF
--- a/config.c
+++ b/config.c
@@ -149,11 +149,12 @@ static int globerr(const char *pathname ATTRIBUTE_UNUSED, int theerr);
 
 static char *isolateLine(char **strt, char **buf, size_t length) {
 	char *endtag, *start, *tmp;
+	const char *max = *buf + length;
 	start = *strt;
 	endtag = start;
-	while (endtag - *buf < length && *endtag != '\n') {
+	while (endtag < max && *endtag != '\n') {
 		endtag++;}
-	if (endtag - *buf > length)
+	if (max < endtag)
 		return NULL;
 	tmp = endtag - 1;
 	while (isspace((unsigned char)*endtag))
@@ -167,16 +168,17 @@ static char *isolateValue(const char *fileName, int lineNum, char *key,
 			char **startPtr, char **buf, size_t length)
 {
     char *chptr = *startPtr;
+    const char *max = *startPtr + length;
 
-    while (chptr - *buf < length && isblank((unsigned char)*chptr))
+    while (chptr < max && isblank((unsigned char)*chptr))
 	chptr++;
-    if (chptr - *buf < length && *chptr == '=') {
+    if (chptr < max && *chptr == '=') {
 	chptr++;
-	while ( chptr - *buf < length && isblank((unsigned char)*chptr))
+	while ( chptr < max && isblank((unsigned char)*chptr))
 	    chptr++;
     }
 
-    if (chptr - *buf < length && *chptr == '\n') {
+    if (chptr < max && *chptr == '\n') {
 		message(MESS_ERROR, "%s:%d argument expected after %s\n",
 			fileName, lineNum, key);
 		return NULL;
@@ -188,14 +190,15 @@ static char *isolateValue(const char *fileName, int lineNum, char *key,
 
 static char *isolateWord(char **strt, char **buf, size_t length) {
 	char *endtag, *start;
+	const char *max = *buf + length;
 	char *key;
 	start = *strt;
-	while (start - *buf < length && isblank((unsigned char)*start))
+	while (start < max && isblank((unsigned char)*start))
 		start++;
 	endtag = start;
-	while (endtag - *buf < length && isalpha((unsigned char)*endtag)) {
+	while (endtag < max && isalpha((unsigned char)*endtag)) {
 		endtag++;}
-	if (endtag - *buf > length)
+	if (max < endtag)
 		return NULL;
 	key = strndup(start, endtag - start);
 	*strt = endtag;
--- a/config.c
+++ b/config.c
@@ -1434,7 +1434,7 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
                                                                            || *start == '~'
 #endif
                                                                            ) {
-				char *key;
+				char *local_key;
 				in_config = 0;
 				if (newlog != defConfig) {
 					message(MESS_ERROR, "%s:%d unexpected log filename\n",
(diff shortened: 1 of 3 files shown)


kerolasa:
Strange that my compiler did not tell about none existing fields in initializer. Oh well, lets see if this amend makes all work.

    diff --git a/logrotate.c b/logrotate.c
    index a4201ae..be29b1d 100644
    --- a/logrotate.c
    +++ b/logrotate.c
    @@ -2470,12 +2470,13 @@ int main(int argc, const char **argv)
     	{"force", 'f', 0, &force, 0, "Force file rotation", NULL},
     	{"mail", 'm', POPT_ARG_STRING, &mailCommand, 0,
     	 "Command to send mail (instead of `" DEFAULT_MAIL_COMMAND "')",
    -	 "command", NULL},
    +	 "command"},
     	{"state", 's', POPT_ARG_STRING, &stateFile, 0,
     	 "Path of state file",
    -	 "statefile", NULL},
    +	 "statefile"},
     	{"verbose", 'v', 0, 0, 'v', "Display messages during rotation", NULL},
    -	{"log", 'l', POPT_ARG_STRING, &logFile, 'l', "Log file or 'syslog' to log to syslog", NULL},
    +	{"log", 'l', POPT_ARG_STRING, &logFile, 'l', "Log file or 'syslog' to log to syslog",
    +	 "logfile"},
     	{"version", '\0', POPT_ARG_NONE, NULL, 'V', "Display version information", NULL},
     	POPT_AUTOHELP { NULL, 0, 0, NULL, 0, NULL, NULL }
         };

logrotate.h
```diff
@@ -45,9 +45,9 @@ struct logInfo {
     char *oldDir;
     enum { ROT_HOURLY, ROT_DAYS, ROT_WEEKLY, ROT_MONTHLY, ROT_YEARLY, ROT_SIZE
             } criterium;
-    unsigned long long threshhold;
-    unsigned long long maxsize;
-    unsigned long long minsize;
+    size_t threshhold;
```
kdudka: This does not look right.  You changed the internal representation and printf conversions to ```size_t``` despite the value is assigned from ```unsigned long long``` and later explicitly converted to ```off_t``` before a comparison.  Looks like just too many conversions for nothing.  Why did you choose ```size_t```?

kdudka:
I have pushed the approved commits to the ```warnings``` branch for a review:
https://github.com/logrotate/logrotate/compare/warnings

config.c
```diff
@@ -142,11 +142,12 @@ static int globerr(const char *pathname __attribute__((__unused__)), int theerr)
 
 static char *isolateLine(char **strt, char **buf, size_t length) {
 	char *endtag, *start, *tmp;
+	const char *max = max = *buf + length;
```
kerolasa: Oh yes, typo indeed. Fixed in updated pull request.

config.c
```diff
@@ -1434,7 +1434,7 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
                                                                            || *start == '~'
 #endif
                                                                            ) {
-				char *key;
+				char *Key;
```
kerolasa: Variable name is changed to local_key.

config.c
```diff
@@ -138,7 +138,7 @@ int tabooCount = 0;
 static int glob_errno = 0;
 
 static int readConfigFile(const char *configFile, struct logInfo *defConfig);
-static int globerr(const char *pathname, int theerr);
+static int globerr(const char *pathname __attribute__((__unused__)), int theerr);
```
kerolasa: Attribute usage is done with conditional definition similar to what one can find from gnulib. I am sure that's good enough compromise.

logrotate.c
```diff
@@ -2464,18 +2464,19 @@ int main(int argc, const char **argv)
 
     struct poptOption options[] = {
 	{"debug", 'd', 0, 0, 'd',
-	 "Don't do anything, just test (implies -v)"},
-	{"force", 'f', 0, &force, 0, "Force file rotation"},
+	 "Don't do anything, just test (implies -v)", NULL},
+	{"force", 'f', 0, &force, 0, "Force file rotation", NULL},
```
kerolasa: Not a problem, but a warning =Wall and -Wextra are in use: missing an initializer. Sample warning is added to commit message.

logrotate.h
kerolasa: You are right, off_t would have been better type with printf using %jd && (intmax_t) cast. Done that way.

config.c
```diff
@@ -34,6 +34,13 @@
 
 #define REALLOC_STEP    10
 
+/* The __unused__ attribute was added in gcc 3.2.7.  */
+#if __GNUC__ >= 3 || (__GNUC__ == 2 && __GNUC_MINOR__ >= 7)
+# define ATTRIBUTE_UNUSED __attribute__((__unused__))
+#else
+# define ATTRIBUTE_UNUSED /* empty */
+#endif
+
```
kdudka: What exactly is the problem with the following one-line fix I proposed?

```
--- a/config.c
+++ b/config.c
@@ -710,6 +710,8 @@ int readAllConfigPaths(const char **paths)

 static int globerr(const char *pathname, int theerr)
 {
+    (void) pathname;
+
     /* A missing directory is not an error, so return 0 */
     if (theerr == ENOTDIR)
         return 0;
```

The fix you are proposing is way more complicated and would work with GCC only anyway.

logrotate.c
```diff
@@ -243,7 +243,7 @@ static void unescape(char *arg)
 #define HASH_SIZE_MIN 64
 static int allocateHash(unsigned int hs)
 {
-	int i;
+	size_t i;
```
kdudka: Why did you choose ```size_t``` here?  Would not ```unsigned int``` lead to fewer implied casts?

logrotate.c
```diff
@@ -1361,14 +1361,14 @@ int prerotateSingleLog(struct logInfo *log, int logNum, struct logState *state,
 	 * Construct the glob pattern corresponding to the date format */
 	dext_str[0] = '\0';
 	if (log->dateformat) {
-		i = j = 0;
+		size_t ii = 0, jj = 0;
```
kdudka: I like moving of the definitions of ```i``` and ```j``` to the innermost block but their rename is unnecessary.  It can be avoided by moving the definition of the remaining ```i``` variable to the innermost block, too.

kdudka:
I have updated the ```warnings``` branch with the latest version, including my review comments:
https://github.com/logrotate/logrotate/compare/warnings

Please let me know if you agree with merging it such.

kerolasa:
Thank you @kdudka git://github.com/logrotate/logrotate.git warnings branch looks good. It would be great to get that merged so that we don't have a branch hanging arond for weeks.

kdudka:
On Sunday, January 08, 2017 09:15:25 Sami Kerola wrote:
> It would be great to get that merged so that we don't have a branch
> hanging arond for weeks.

Sure.  Thanks for the contribution!

PR #102, opened 2017-01-15
Remove config.h in favor of configure.ac

schaal:
This moves the values defined in config.h to configure.ac and adds configure options to modify these values.

I tried to map the host values from config.h to the host values from AC_CANONICAL_HOST (e.g. `__NetBSD__` -> `*-netbsd*`), but am not sure if they exactly map each other.


configure.ac
```diff
@@ -57,30 +85,88 @@ AC_ARG_WITH([state-file-path],
...
-  [
-    STATE_FILE_PATH="/var/lib/logrotate.status"
   ])
 AC_DEFINE_UNQUOTED([STATEFILE], "$STATE_FILE_PATH")
 AC_SUBST(STATE_FILE_PATH)
 
 AC_ARG_WITH([default-mail-command],
   AC_HELP_STRING([--with-default-mail-command=COMMAND],
-                 [default mail command (based on config.h if not given)]),
+                 [default mail command (default: /bin/mail -s)]),
```
glensc: i think you can use `$DEFAULT_MAIL_COMMAND` variable here

configure.ac
```diff
@@ -57,30 +85,88 @@ AC_ARG_WITH([state-file-path],
...
-        AC_SUBST(DEFAULT_MAIL_COMMAND)
         ;;
     esac
   ])
+AC_DEFINE_UNQUOTED([DEFAULT_MAIL_COMMAND], "$DEFAULT_MAIL_COMMAND")
+AC_SUBST(DEFAULT_MAIL_COMMAND)
+
+AC_ARG_WITH([compress-command],
+  AC_HELP_STRING([--with-compress-command=COMMAND],
+                 [compress command (default: /bin/gzip)]),
```
glensc: same here, try `$COMPRESS_COMMAND`

configure.ac
```diff
@@ -57,30 +85,88 @@ AC_ARG_WITH([state-file-path],
...
+AC_DEFINE_UNQUOTED([UNCOMPRESS_COMMAND], "$UNCOMPRESS_COMMAND")
+AC_SUBST(UNCOMPRESS_COMMAND)
+
+AC_ARG_WITH([compress-extension],
+  AC_HELP_STRING([--with-compress-extension=EXTENSION],
+                 [compress extension (default: .gz)]),
+  [
+    case "$withval" in
+      yes|no)
+        AC_MSG_ERROR([--with-compress-extension=EXTENSION requires an extension to be specified])
```
glensc: i would allow `yes`, and `no` extensions here :)

glensc: however, perhaps just test that the value begins with dot:

```
COMPRESS_EXT=".gz"
```

so that `--with-compress-extension=gz` is invalid, but `--with-compress-extension=.gz` is okay

configure.ac
schaal: I tried this, but it doesn't get expanded

configure.ac
schaal: OK, changed it to only accept values starting with dot

configure.ac
kdudka: Should we remove the ```-s``` then?

Because ```-s``` is used by default for ```*-netbsd*``` only if I understand it correctly.

schaal: Maybe it would be better not calling it default, but instead e.g., like `default mail command (e.g. /bin/mail -s)`, as the default is not known yet when the help text is shown

kerolasa:
FYI I created moment ago pull 103, that takes different approach to deal mail command. See link below for details.

https://github.com/logrotate/logrotate/pull/103/commits/79a5c5a572e441eb1c15feb23cc68717ce967f43

kdudka:
@kerolasa Since this PR is already approved, I would prefer to push this one first and rebase your configure patch on top of this one, picking only the changes that will still be needed.

@glensc What is your opinion on this?

glensc:
same point here, this should be merged first

PR #108, opened 2017-03-06
Fixes GCC sentinel error.

stevenwilliamson:
The last argument for sentinel value functions should be a NULL pointer not just a NULL value.
Otherwise GCC 4.x will error with "error: missing sentinel in function call [-Werror=format=]"

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -440,7 +440,7 @@ static int runScript(struct logInfo *log, char *logfn, char *script)
 				DOEXIT(1);
 			}
 		}
-		execl("/bin/sh", "sh", "-c", script, "logrotate_script", logfn, NULL);
+		execl("/bin/sh", "sh", "-c", script, "logrotate_script", logfn, (char *)NULL);
 		DOEXIT(1);
 	}
 
@@ -800,7 +800,7 @@ static int mailLog(struct logInfo *log, char *logFile, char *mailComm,
 				DOEXIT(1);
 			}
 
-			execlp(uncompressCommand, uncompressCommand, NULL);
+			execlp(uncompressCommand, uncompressCommand, (char *)NULL);
 			DOEXIT(1);
 		}
 


kdudka:
This looks suspicious to me.  I am able to compile logrotate using ```gcc-4.9.4``` without any warnings.  The explicit type conversion looks unnecessary.  On my system ```NULL``` is defined as ```((void *)0)``` and any compliant C compiler should convert ```(void *)``` to ```(char *)``` implicitly.  Please attach the full build log and some steps to reproduce the error.

stevenwilliamson:
Missed this comment sorry for the delay. Build log attached, this is from pkgsrc on SmartOS.
I guess this is a portability issue NULL might not always be defined as (void *) ?

[log.txt](https://github.com/logrotate/logrotate/files/902851/log.txt)

kdudka:
Could you please run the following command?
```
gcc -std=gnu99 -DPACKAGE_NAME=\"logrotate\" -DPACKAGE_TARNAME=\"logrotate\" -DPACKAGE_VERSION=\"3.10.0\" -DPACKAGE_STRING=\"logrotate\ 3.10.0\" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -DPACKAGE=\"logrotate\" -DVERSION=\"3.10.0\" -D_GNU_SOURCE=1 -DSTDC_HEADERS=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_MEMORY_H=1 -DHAVE_STRINGS_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_UNISTD_H=1 -DHAVE_STRUCT_STAT_ST_BLKSIZE=1 -DHAVE_ST_BLKSIZE=1 -DHAVE_STRUCT_STAT_ST_BLOCKS=1 -DHAVE_ST_BLOCKS=1 -DHAVE_LIBPOPT=1 -DHAVE_STRPTIME=1 -DHAVE_QSORT=1 -DHAVE_FORK=1 -DHAVE_VFORK=1 -DHAVE_MADVISE=1 -DHAVE_VSYSLOG=1 -I.   -I/opt/local/include -I/usr/include -Wall -Werror -O2 -pipe -O2 -gdwarf-2 -I/opt/local/include -I/usr/include -MT logrotate.o -MD -MP -MF .deps/logrotate.Tpo -E -o logrotate.i logrotate.c
```
... and attach the resulting ```logrotate.i``` file?

kdudka:
@stevenwilliamson ping?

The-42:
I recall having a similar problem with gcc 4.9.4 some time ago, it seems to depend a bit on the glibc version you're working with. More recent versions and any gcc 5+ appear to be less picky about the terminator.

However, both POSIX and `man execl` state clearly: 

> The list of arguments must be terminated by a null pointer,  and,  since  these  are  variadic  functions,  this pointer must be cast (char *) NULL.

Since `NULL` can indeed be defined as `0` (ANSI C, I think) instead of `(void *) 0`, this patch is technically correct.

:+1:

kdudka:
@The-42 thanks for the reference!  I will push it as it is.
@stevenwilliamson Thanks for the patch!

PR #114, opened 2017-04-06
SECURITY: createolddir: Create old directory as unprivileged user

fumiyas:
Fix insecure old directory creation.

If parent directory of olddir is writeable by an attacker (non-root),
the attacker can override mode, owner and group of arbitrary files.


kdudka:
Looks good to me.  Thanks for the contribution!

PR #116, opened 2017-04-10
2017wk15 various small fixes

kerolasa:
This pull request contains fixes to various static analyzers runs (smatch, cppcheck, llvm scan-build). There are also updates to documentation. If you look commit timestamps you can see that these were done sometime ago - waiting previous autotools pull to be processed. After these changes I do not have anything in my backlog, and it might be good time to cut a release.

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -1914,7 +1914,12 @@ static int rotateLogSet(struct logInfo *log, int force)
     struct logState **state;
     struct logNames **rotNames;
 
-    logHasErrors = alloca(log->numFiles * sizeof(int));
+    logHasErrors = calloc(log->numFiles, sizeof(int));
+    if (!logHasErrors) {
+        message(MESS_ERROR, "could not allocate memory for "
+                "logHasErrors\n");
+        return 1;
+    }
     message(MESS_DEBUG, "\nrotating pattern: %s ", log->pattern);
     if (force) {
         message(MESS_DEBUG, "forced from command line ");
@@ -1974,11 +1979,13 @@ static int rotateLogSet(struct logInfo *log, int force)
 
 	if (log->numFiles == 0) {
 		message(MESS_DEBUG, "No logs found. Rotation not needed.\n");
+		free(logHasErrors);
 		return 0;
 	}
 
 	if (log->flags & LOG_FLAG_SU) {
 		if (switch_user(log->suUid, log->suGid) != 0) {
+			free(logHasErrors);
 			return 1;
 		}
 	}
@@ -2005,10 +2012,12 @@ static int rotateLogSet(struct logInfo *log, int force)
 		hasErrors = 1;
 		if (log->flags & LOG_FLAG_SU) {
 			if (switch_user_back() != 0) {
+				free(logHasErrors);
 				return 1;
 			}
 		}
 		/* finish early, firstaction failed, affects all logs in set */
+		free(logHasErrors);
 		return hasErrors;
 	    }
 	}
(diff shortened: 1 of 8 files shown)


logrotate.c
```diff
@@ -552,6 +553,7 @@ static int createOutputFile(char *fileName, int flags, struct stat *sb,
 		return -1;
 		}
 	}
+#endif
```
kdudka: The proposed change does not seem to be correct.  The condition of the if block you commented out is ```(!acl_set || force_mode)``` and ```force_mode``` can be true regardless of ```WITH_ACL```.

logrotate.c
```diff
@@ -1914,7 +1914,12 @@ static int rotateLogSet(struct logInfo *log, int force)
     struct logState **state;
     struct logNames **rotNames;
 
-    logHasErrors = alloca(log->numFiles * sizeof(int));
+    logHasErrors = calloc(sizeof(int), log->numFiles);
```
kdudka: Looks good.  The swap of operands is unnecessary though.  The prototype of ```calloc()``` is:
```
void *calloc(size_t nmemb, size_t size);
````
No need to resubmit, of course.

File: README.md [hunk 3]
```diff
@@ -75,10 +75,10 @@ Install dependencies for Fedora/CentOS systems:
 yum install make autoconf automake libtool popt-devel xz
 ```
 
-Compilation (`autoreconf` is optional if you obtained source from tarball):
+Compilation (`./autogen.sh` is optional if you obtained source from tarball):
 ```
 cd logrotate-X.Y.Z
-autoreconf -fiv
+./autogen.sh
```
kdudka: What is the advantage of recommending to run ```autogen.sh``` instead of ```autoreconf```?

```autogen.sh``` does not provide any options, neither has its own man page.  When something goes wrong, one would have to debug yet another layer for no apparent benefit.  I thought that ```autogen.sh``` was just a wrapper for compatibility with legacy build scripts.

kdudka:
Looks good.  Besides the minor issues mentioned above, this seems to be ready for merge.

kdudka:
I have prepared it for merge in the ```kerolasa``` branch:
https://github.com/logrotate/logrotate/compare/kerolasa

I was only not sure with commit a590ff66bf43eacfc49551964000e08ef0ffc492 because I do not know which warning of which tool we are fixing by that.

@kerolasa could you please have a look?

logrotate.c
kerolasa: Oh, I missed that. Thank you for careful review.

logrotate.c
kerolasa: No idea what I did there. Thanks for taking care of that.

README.md
kerolasa: My thinking was (and is) that it's best to recommend to use facilities provided by project. One day autogen.sh might contain more commands, lets say logrotate will be translated and gettext stuff has to be taken care, and then it is nice if people are used ot do what is considered the right way of bootstrapping the project.  But right now the argument in behalf of autogen.sh is not very strong, so it's fine to say autoreconf does what is needed.

kerolasa:
I had had a look of origin/kerolasa branch and it looks good. Lets get the thing merged.

kerolasa:
BTW if you merge from my 2017wk15 branch then origin will get exactly the contents of origin/kerolasa commits + all commits signed with my gpg key 8ED396E37E38D471A00530D3A9553245FDE9B739.

kdudka:
Pushed.  Thanks!

PR #119, opened 2017-04-21
fix unused function error when building without ACL

The-42:
Since -Werror is set, building without ACL makes the then unused
is_acl_well_supported function halt the build. Guard it as it is done
with every call of it.

Signed-off-by: Bert van Hall <bert.vanhall@avionic-design.de>

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -442,6 +442,7 @@ static int runScript(struct logInfo *log, char *logfn, char *script)
 	return rc;
 }
 
+#ifdef WITH_ACL
 static int is_acl_well_supported(int err)
 {
 	switch (err) {
@@ -454,6 +455,7 @@ static int is_acl_well_supported(int err)
 		return 1;
 	}
 }
+#endif /* WITH_ACL */
 
 static int createOutputFile(char *fileName, int flags, struct stat *sb,
 			    acl_type acl, int force_mode)


kdudka:
Looks good to me.  Thanks!

kdudka:
Btw. I do not think we should enforce ```-Werror``` in production builds.  @glensc What do you think?

The-42:
My 2 cents: Give maintainers a chance to disable -Werror.

Could look like this in configure.ac:
```autoconf
AC_ARG_ENABLE([werror],
        [AS_HELP_STRING([--disable-werror],
                [Do not treat warnings as errors (default: warnings are errors)])],
                [enable_werror="$enableval"],
                [enable_werror=yes])
AS_IF([test "x$enable_werror" = "xyes"], [AX_APPEND_COMPILE_FLAGS([-Werror])])
```
(uses the [ax_append_compile_flags](https://www.gnu.org/software/autoconf-archive/ax_append_compile_flags.html) macro)

glensc:
@kdudka agreed. we can add `--enable-werror` in in `.travis.yml`

kdudka:
@The-42 @glensc I have submitted it as #123.  Could you please review?

pickfire:
Is there going to be logrotate-3.12.1 or the maintainers need to patch it themselves?

kdudka:
Good point.  I did not realize that this issue had been introduced in ```logrotate-3.12.0```.

I will release ```logrotate-3.12.2``` today to get this fixed...

PR #123, opened 2017-04-25
configure --enable-werror can be used to re-enable -Werror

kdudka:
The -Werror compiler flag is no longer used by default.


kdudka:
Thanks for review!

PR #126, opened 2017-05-19
config.c / manual: list taboo extensions in alphabetical order

kerolasa:
Alphabetical order helps users to read what is in the list.  In config.c
horizontal list is converted to vertical to allow version control to track
each extension separately.  Vertical ordered lists are also easy to cross
check, and indeed that helped noticing manual page had missed ,v extension
that was probably added to avoid RCS files being mangled by the logrotate.

### DIFF
--- a/config.c
+++ b/config.c
@@ -126,10 +126,22 @@ enum {
 	STATE_ERROR = 64,
 };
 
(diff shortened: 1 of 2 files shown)


kdudka:
Looks good to me.  Thanks!

Issue #22, opened 2016-02-23
logrotate exits with success even in case of errors


davide125:
Logrotate will exit with 0 even when it failed to rotate because of an error (e.g. invalid config file). In this case it should exit with 1 or some more specific exit code, but definitely not success. Repro example:

```
# cat logrotate.conf 
weekly
rotate 4
create
dateext
/var/log/wtmp {
    monthly
    create 0664 root utmp
    minsize 1M
    rotate 1
}
/var/log/wtmp {
    daily
    create 0664 root utmp
    minsize 1M
    rotate 1
}
# /usr/sbin/logrotate logrotate.conf
error: logrotate.conf:22 duplicate log entry for /var/log/wtmp
# echo $?
0
# logrotate --version
logrotate 3.9.2
```

hanzz:
Logrotate should return 1 for most errors. For this particular one, logrotate still tries to rotate the file according to the first rotation defition and therefore it is not considered to be a fatal error, so it just prints error message, but continues with rotation and exits with 0.

The error message could be probably changed to warning or something like that, so it would be clear the rotation still continues.

gdubicki:
It's a good idea to change the message from error to warning. "error" is misleading.

kdudka:
@gdubicki Which version of logrotate are you using?

I believe that the latest release of logrotate includes two fixes, which are closely related:
1. it propagates the error codes while reading the configuration: 6a75cdeab61a29ea99d49a85461f597c1d8d055c
2. it recovers from errors while reading the configuration: 56598fd1e9338b45ba3e1bda1b91522e75b40e06 and 74b788f790990d5958314df5f908a6fc5eaeccdd

kdudka:
The original example now returns non-zero exit code, so I close this as fixed.  Please report separate issues if anything else is needed.

PR #131, opened 2017-06-27
Log it as normal as it doesn't stop logrotate run

gdubicki:
Fixes https://github.com/logrotate/logrotate/issues/22


kdudka:
It is currently handled as an error, although not a fatal error.  Please see my comment at #22.

kdudka:
I believe the patch is no longer needed.  Please update to the latest release of logrotate and retest.

PR #132, opened 2017-07-04
run each test-case separately, possibly in parallel

kdudka:


scarabeusiv:
Seems the tests are now failing if one sets different compression method than gz.
Mostly in all locations, it operates with hardcoded gzip gunzip and suffix to be .gz.

It should be a separate issue I know, but since you are already rewriting the tests..

kdudka:
On Friday, July 07, 2017 02:02:36 Tomáš Chvátal wrote:
> It should be a separate issue I know

Exactly.  Unless it worked before my changes, it is a separate issue.

> but since you are already rewriting the tests..

I am not rewriting them, just porting them to automake.  The code of each 
single test stays unchanged unless there was a bug preventing the test from 
being ported.

Please open a separate PR once this one is merged.

kdudka:
I am going to merge this in two weeks in case there is no feedback till then.

PR #137, opened 2017-08-10
tests: specify (un)compress utility explicitly

kdudka:
... to prevent test failures if a custom compression utility is set
using the ```--with-compress-command```, ```--with-uncompress-command```, and
```--with-compress-extension``` build-time options.

### DIFF
--- a/test/test-config.10.in
+++ b/test/test-config.10.in
@@ -4,6 +4,9 @@ create
     daily
     rotate 3
     compress
+    compresscmd gzip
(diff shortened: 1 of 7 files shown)


kdudka:
@scarabeusiv Could you please check if this patch makes the test-suite pass in your environment?

scarabeusiv:
Works in OBS fine as far as I can see.

kdudka:
Thanks for review!  Pushing to master...

PR #138, opened 2017-08-12
2017wk32: couple test improvements

kerolasa:
I see that autotools parallel tests are made to work. Great job with that. Now when the test suite is a lot easier to work with here are couple improvements to them.

### DIFF
--- a/test/test-0018.sh
+++ b/test/test-0018.sh
@@ -5,34 +5,38 @@
 cleanup 18
 
 # ------------------------------- Test 18 ------------------------------------
-preptest test.log 18 1
+needle="$(uuidgen -r)"
+
+preptest "$needle.log" 18 1
 $RLR test-config.18 -l syslog --force
 
 checkoutput <<EOF
-test.log 0
-test.log.1.Z 1 zero
+$needle.log 0
+$needle.log.1.Z 1 zero
 EOF
 
+rm -f "$needle.log" "$needle.log.1.Z"
+
 (echo "gzip -f -9") | diff -u - compress-args
-egrep -q '^LOGROTATE_COMPRESSED_FILENAME=.+/test.log.1$' compress-env
+grep -E -q "^LOGROTATE_COMPRESSED_FILENAME=.+/$needle.log.1$" compress-env
 if [ $? != 0 ]; then
-      echo "LOGROTATE_COMPRESSED_FILENAME environment variable not found."
+      echo "LOGROTATE_COMPRESSED_FILENAME $needle environment variable not found."
       cat compress-env
       exit 3
 fi
 
 SYSLOG_TESTS=0
-logger syslog_test 2>/dev/null
+logger "logrotate test suite $needle" 2>/dev/null
 if [ $? = 0 ]; then
-	journalctl -n 50 2>/dev/null | grep syslog_test 2>/dev/null >/dev/null
+	journalctl -b -n 50 _COMM='logger' 2>/dev/null | grep "$needle" 2>/dev/null >/dev/null
 	if [ $? = 0 ]; then
 		SYSLOG_TESTS=1
 	fi
 fi
 if [ $SYSLOG_TESTS = 1 ]; then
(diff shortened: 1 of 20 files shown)


test/test-common.sh
```diff
@@ -143,7 +143,7 @@ checkoutput() {
 	    echo expected: \'$expected\'
 	    exit 2
 	fi
-	echo "$config_crc" | md5sum -c - 2>&1 > /dev/null
+	echo "$config_crc" | md5sum -c - >/dev/null 2>&1
```
kdudka: While I agree it as poorly coded, I do not like the fix.  Why should we hide the eventual diagnostic message from the user/developer in the first place?

If something goes wrong, the user should know about it. The correct fix would IMO look such:
```
--- a/test/test-common.sh
+++ b/test/test-common.sh
@@ -143,7 +143,7 @@ checkoutput() {
        echo expected: \'$expected\'
        exit 2
    fi
-   echo "$config_crc" | md5sum -c - 2>&1 > /dev/null
+   echo "$config_crc" | md5sum -c - > /dev/null
    if [ $? != 0 ]; then
        echo "config file $output has been altered: MD5 sum mismatch"
        exit 3
```

test/test-0018.sh
```diff
@@ -5,34 +5,38 @@
 cleanup 18
 
 # ------------------------------- Test 18 ------------------------------------
-preptest test.log 18 1
+needle="$(uuidgen -r)"
```
kdudka: On my system ```uuidgen``` is installed by the util-linux package?  Is it available on non-linux platforms (e.g. cygwin)?  If not, we should not use it unconditionally.  Why should any files survive across test runs?  Is not the proper fix to just clean files properly before running the test?

test/test-common.sh
kerolasa: The stderr output is made visible in 4e920978db87d0b291e25fe9262b55b5c442c811

test/test-0018.sh
kerolasa: I don't think we need to be too worried about availability of `uuidgen` when `journalctl` is used in the same script. To be honest `test-0018.sh` should be flagged as linux only.

Survival of results is about journal entries that are system wide and span over separate test runs, sessions, and so on. In short the grep in test needs to have something uniq to match for each check run, else results can be ambiguous.

kdudka: How are you confirming that ```journalctl``` is unconditionally used?

The test passes just fine on my system where ```journalctl``` does not exist.  Not that I understand what the test actually checks but we need to fix it to exit with 77 if its unconditional requirements are not met.  As it works now, either the ```journalctl``` availability is optional, or the test is badly broken.

kerolasa: > How are you confirming that journalctl is unconditionally used?

Not exactly. Lack of journalctl causes SYSLOG_TESTS value to remain zero, and the finding a message part of the test is not done. In effect, if I understand the idea of this test correctly, the logrotate syslogging is only verified on systems that has working journal. Other systems simply skip the verification, which is not great but we can perhaps tolerate that - at least the check can work somewhere. But as I mentioned earlier journal entries are global, that is not session or user specific, one has to be careful not to mix a check payload from one run with an another. Hence the change I wrote.

kdudka: OK.  So unless you want to split the test into two separate tests (one generic and one Linux specific), the following patch should avoid the test-suite failure on systems without ```uuidgen```:
```
--- a/test/test-0018.sh
+++ b/test/test-0018.sh
@@ -2,6 +2,9 @@

 . ./test-common.sh

+# uuidgen is required by this test, availability of journalctl is optional
+uuidgen --version || exit 77
+
 cleanup 18

 # ------------------------------- Test 18 ------------------------------------
```
Are you fine with this fixup?

kdudka: @kerolasa ping?

kdudka: This seems to get staled.  I will push everything except this commit to move things forward.  Please open a separate pull request for the changes to ```test-0018.sh``` if you still want to get them in...

Issue #141, opened 2017-09-01
Systemd hardening options


cgzones:
Currently I am testing logrotate with the following additional systemd service options:

```
ProtectSystem=full
ProtectHome=true
PrivateDevices=true
ProtectKernelTunables=true
ProtectControlGroups=true
ProtectKernelModules=true
PrivateNetwork=true
```

Could some of these get included?

kdudka:
Any systemd expert around to review this?  Could it break any commonly used lgoratate configurations?

Please open a pull request for this.  From your description it is not clear where exactly these lines should go...

glensc:
#143 got merged, so this can be closed i believe

PR #143, opened 2017-09-04
logrotate.service: add hardening options

cgzones:
implements #141

### DIFF
--- a/examples/logrotate.service
+++ b/examples/logrotate.service
@@ -6,6 +6,21 @@ ConditionACPower=true
 [Service]
 Type=oneshot
 ExecStart=/usr/sbin/logrotate /etc/logrotate.conf
+
+# performance options
 Nice=19
 IOSchedulingClass=best-effort
 IOSchedulingPriority=7
+
+# hardening options
+#  details: https://www.freedesktop.org/software/systemd/man/systemd.exec.html
+#  no ProtectHome for userdir logs
+#  no PrivateNetwork for mail deliviery
+MemoryDenyWriteExecute=true
+PrivateDevices=true
+PrivateTmp=true
+ProtectControlGroups=true
+ProtectKernelModules=true
+ProtectKernelTunables=true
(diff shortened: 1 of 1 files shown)


glensc:
could you provide link to documentation for each directive?

from name guessing i do see some issues:

- ProtectHome=true

may break logrotate configurations that rotate files in user's dir (example /home/tomcat/logs). but then maybe it shouldn't be rotated using system service?

- PrivateNetwork=true

this may break logrotate configurations to invoke `/bin/mail` to deliver mails, unless the `mail` command only queues locally. and of course all `action` blocks can contain basically anything may need network interaction. i guess it would be still fine if these settings are properly documented so people know where to look into and how to resolve issues.

cgzones:
The directives can be found at: https://www.freedesktop.org/software/systemd/man/systemd.exec.html

You concerns might be correct, so an opt-in might be better.

cgzones:
`ProtectKernelTunables=true` breaks SELinux support, cause it needs a writeable `/sys/fs/selinux`

kdudka:
@cgzones Please let me know if you are fine with the requested changes.  I can squash them in for you if you agree.  No need to resubmit...

cgzones:
since systemd v235 ProtectKernelTunables does not break SELinux anymore

kdudka:
On Saturday, October 21, 2017 9:39:42 PM CET cgzones wrote:
> since systemd v235 ProtectKernelTunables does not break SELinux anymore

That is good.  But we cannot safely assume that users of logrotate run the 
latest version of systemd.  It is IMO better to keep the list of hardening 
options conservative than possibly break existing setups.

kdudka:
@cgzones Please let me know if you are fine with the requested changes. I can squash them in for you if you agree. No need to resubmit...

cgzones:
I am fine with the changes

kdudka:
OK, merging now... Thanks!

PR #145, opened 2017-09-19
add compile time settings into version output

cgzones:
example output:

$ logrotate --version
logrotate 3.12.3

    Default mail command: /bin/mail
    Compress command:     /bin/gzip
    Uncompress command:   /bin/gunzip
    Compress extension:   .gz
    State file path:      /var/lib/logrotate.status
    ACL support:          yes
    SELinux support:      yes

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -2599,22 +2599,22 @@ int main(int argc, const char **argv)
 		}
 		break;
 	case 'V':
-	    fprintf(stderr, "logrotate %s\n", VERSION);
-	    fprintf(stderr, "\n");
-	    fprintf(stderr, "    Default mail command:       %s\n", DEFAULT_MAIL_COMMAND);
-	    fprintf(stderr, "    Default compress command:   %s\n", COMPRESS_COMMAND);
-	    fprintf(stderr, "    Default uncompress command: %s\n", UNCOMPRESS_COMMAND);
-	    fprintf(stderr, "    Default compress extension: %s\n", COMPRESS_EXT);
-	    fprintf(stderr, "    Default state file path:    %s\n", STATEFILE);
+	    fprintf(stdout, "logrotate %s\n", VERSION);
+	    fprintf(stdout, "\n");
+	    fprintf(stdout, "    Default mail command:       %s\n", DEFAULT_MAIL_COMMAND);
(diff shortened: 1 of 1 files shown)


glensc:
what has always bothered me, that tools use `stderr` for intended messages (not errors), so they are not easily pipeable. imho version output should go to `stdout`.

logrotate.c
```diff
@@ -2600,6 +2600,22 @@ int main(int argc, const char **argv)
 		break;
 	case 'V':
 	    fprintf(stderr, "logrotate %s\n", VERSION);
+	    fprintf(stderr, "\n");
+	    fprintf(stderr, "    Default mail command: %s\n", DEFAULT_MAIL_COMMAND);
+	    fprintf(stderr, "    Compress command:     %s\n", COMPRESS_COMMAND);
+	    fprintf(stderr, "    Uncompress command:   %s\n", UNCOMPRESS_COMMAND);
+	    fprintf(stderr, "    Compress extension:   %s\n", COMPRESS_EXT);
+	    fprintf(stderr, "    State file path:      %s\n", STATEFILE);
```
kdudka: I think we should use the ```Default ``` prefix in all the lines above.  This is especially important for the state file, which can be overridden by a command-line option (and certain distros use the option in default cron jobs to override logrotate's default) but I would put it everywhere to stay consistent.

kdudka:
@glensc Agreed.  Are you proposing to change the output stream for ```--version``` along with this PR?

In any case, such a change needs to be properly advertised in release notes because it is a backward incompatible change.

glensc:
@kdudka yes, change in this PR, as depending how `--version` output is parsed by external scripts, it could break scripts anyway.

for example `logrotate --version 2>&1 | awk '{print $2}'`.

kdudka:
@cgzones Are you fine with including the change of output stream in this PR?

kdudka:
Can I abbreviate ```fprintf(stdout, ...)``` as ```printf(...)```?

You do not need to resbumit it, of course...

cgzones:
> Can I abbreviate fprintf(stdout, ...) as printf(...)?

yes, of course

kdudka:
OK, pushing now... Thanks!

PR #146, opened 2017-09-19
fix spelling, found by scan-build

cgzones:
logrotate.c:251:15: warning: Result of 'malloc' is converted to a pointer of type 'struct logStates', which is incompatible with sizeof operand type 'struct logState'

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -248,7 +248,7 @@ static int allocateHash(unsigned int hs)
 	}
 
 	for (i = 0; i < hs; i++) {
-		states[i] = malloc(sizeof(struct logState));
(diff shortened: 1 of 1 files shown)


kdudka:
Your fix is correct.  But the currently chosen identifiers ```struct logState``` vs. ```struct logStates``` are really unfortunate.  I propose to take two actions to prevent bugs like this in the future:

1. Rename ```struct logStates``` to ```struct logStateList```.
2. Write:
```
states[i] = malloc(sizeof states[0]);
```
... instead of:
```
states[i] = malloc(sizeof(struct logState));
```

cgzones:
see #149

PR #148, opened 2017-09-20
writeState(): call fflush() before fsync()

rantala:
Flush the libc buffers before calling fsync().

Otherwise the file content may not yet have been written to the kernel,
and fsync() will not help in ensuring that the file data is written to
disk.


kdudka:
Looks good.  Thanks for the patch!

Issue #150, opened 2017-09-21
logrotate always sends -s to the mail command


gpakosz:
Hello while debugging a problem with logrotate I noticed the following: https://github.com/logrotate/logrotate/blob/8523cb352c629a029b27cbf6be110bfe1703237b/logrotate.c#L796

logrotate always sends `-s` to the mail command

However the documentation states:
>        -m, --mail <command>
>              Tells logrotate which command to use when mailing logs. This command should accept two arguments: 1) the subject of the message, and 2) the recipient. The com-
>              mand must then read a message on standard input and mail it to the recipient. The default mail command is /bin/mail -s.
See https://github.com/logrotate/logrotate/blob/8523cb352c629a029b27cbf6be110bfe1703237b/logrotate.8.in#L66-L74

I see two possible resolutions:
  - either put `-s` in `DEFAULT_MAIL_COMMAND`
  - or update the documentation stating the subject is given with `-s`.

Best.

kdudka:
I would go for option 2 (update the documentation stating that subject is given with ```-s```).  Could you open a pull request for that?

Note that the documentation text you quoted is no longer produced by the current version of logrotate.  The trailing ```-s``` was removed in [3.11.0-18-gad5ca64](https://github.com/logrotate/logrotate/commit/ad5ca64349545799cc0783819ba215ffc698b693).

gpakosz:
Ah yeah the documentation I quoted comes from Debian Stretch
```
logrotate 3.11.0 - Copyright (C) 1995-2001 Red Hat, Inc.
This may be freely redistributed under the terms of the GNU Public License
```

gpakosz:
`configure.ac` seems to be inconsistent when defining `DEFAULT_MAIL_COMMAND`:
https://github.com/logrotate/logrotate/blob/68fee20eeff3152ee3e7d176f95fcad53c456c2d/configure.ac#L74

kdudka:
Good point.  It would not work anyway if we gave ```"/usr/bin/mail -s"``` as the file argument to ```execvp()```.  All the ``` -s``` suffixes in ```configure.ac``` need to be removed.

gpakosz:
Right, will do in an upcoming PR.

kdudka:
Resolved by #152

PR #152, opened 2017-09-26
Updated DEFAULT_MAIL_COMMAND in configure.ac and updated man page

gpakosz:
As discussed in #150,

This PR contains 2 commits:
  - a commit that removes `-s` from `DEFAULT_MAIL_COMMAND` definition and help string in `configure.ac`
  - a commit that tentatively rephrases the `logrotate.8.in` man page template


kdudka:
Pushing now... Thanks!

PR #158, opened 2017-10-27
Add automake to Debian dependency list

MrBerg:
At least on my Mint installation, I had to install automake to get autoreconf working, so this might be good to add to the installation instructions.


kdudka:
The patch looks good.  However, while reading the instructions, I realized that we should probably mention that ```automake```, ```autoconf```, and ```libtool``` are necessary only when building a git checkout.  They are not needed when building a release tarball.

PR #159, opened 2017-11-22
add hourago option

torgiren:
Add datehourago option that is similar to dateyesterday, but useful with hourly rotate logs.
I fixes filenames after rotate, that match do hour in file

### DIFF
--- a/config.c
+++ b/config.c
@@ -1063,6 +1063,8 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
 					newlog->flags &= ~LOG_FLAG_DATEEXT;
 				} else if (!strcmp(key, "dateyesterday")) {
 					newlog->flags |= LOG_FLAG_DATEYESTERDAY;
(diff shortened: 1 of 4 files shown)


kdudka:
I see that this directive is similar to the already existing ```dateyesterday``` directive but I was not able to find the reasoning behind introducing the ```dateyesterday``` directive and have never used it myself.  Could you please show us some examples where it makes sense to use (or not to use) the ```datehourago``` directive?

torgiren:
Example:
Tomcat, log catalina.out
date: 2017-11-22 19:09
entries in log have date ex:: 2017-11-22 19:09
at 20:00, log is rotated and (without datehourago) renamed to: catalina.out-2017-11-22-20,
when I expect to rename it to catalina.out-2017-11-22-19 because logs in this file are from hour 19, not 20

kdudka:
Thanks!  I think it makes sense and I will merge it in a week or so unless anybody else comments on this.

PR #162, opened 2017-12-15
added .dpkg-tmp to defTabooExts

fedordikarev:
As we have issues running logrotate during package upgrades:
reading config file project-api
reading config file project-api.dpkg-tmp
error: project-api.dpkg-tmp:1 duplicate log entry for /var/log/project/api.log


kdudka:
Thanks!  Merging now...

PR #166, opened 2018-01-18
Removed check for user execute bit

thehomerepot:
The user execute bit being set on config files was preventing logrotate from utilizing said config. No reason for this restriction.

### DIFF
--- a/config.c
+++ b/config.c
@@ -935,9 +935,9 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
 	}
 
 	if (getuid() == ROOT_UID) {
-		if ((sb.st_mode & 07533) != 0400) {
+		if ((sb.st_mode & 07433) != 0400) {
 			message(MESS_ERROR,
-				"Ignoring %s because of bad file mode - must be 0644 or 0444.\n",
+				"Ignoring %s because of bad file mode - must not be u-r g+wx o+wx.\n",
 				configFile);
 			close(fd);
 			return 0;


kdudka:
Sounds like a reason to fix permissions on the config file, not a reason to paralyze the test.  Having logrotate config files executable does not make any sense to me...

thehomerepot:
Does allowing execute permissions for the owner paralyze the test? If so, what are you preventing by blocking the execute bit from being set for owner? I think there are 2 sides of the coin to view this from.

1. Execute bit for owner isn't necessary for the config file
2. It isn't necessary for logrotate to block the execute bit for owner.

Neither side of the coin actually matter. Arguments can be mode both ways. As a sysadmin, at the end of the day I'd prefer tools not to be restrictive when those restrictions aren't necessary.

kdudka:
On Friday, January 19, 2018 4:16:55 PM CET thehomerepot wrote:
> Does allowing execute permissions for the owner paralyze the test? If so,
> what are you preventing by blocking the execute bit from being set for
> owner?

If you have a text file with executable bit set, it can be executed as a shell 
script, either accidentally by a user, or even by an attacker.  logrotate's 
config files are not intended for execution.  They are intended to be _read_ 
by logrotate.

> I think there are 2 sides of the coin to view this from.
> 
> 1. Execute bit for owner isn't necessary for the config file
> 2. It isn't necessary for logrotate to block the execute bit for owner.
>
> Neither side of the coin actually matter. Arguments can be mode both ways.
> As a sysadmin, at the end of the day I'd prefer tools not to be restrictive
> when those restrictions aren't necessary.

Unless you show us a valid setup that needs executable bit set on logrotate's 
config files, I do not think that weakening the test is a reasonable thing to 
do.

thehomerepot:
> If you have a text file with executable bit set, it can be executed as a shell
script, either accidentally by a user, or even by an attacker.  logrotate's
config files are not intended for execution.  They are intended to be _read_
by logrotate.

In this situation, the attacker would need root access. If they have that, none of this matters. So at that point, what harm does having the execute bit set have?

> Unless you show us a valid setup that needs executable bit set on logrotate's
config files, I do not think that weakening the test is a reasonable thing to
do.

If this is the stance, I don't understand why you don't restrict permissions to 600? Why does group or other need any access to these config files at all? Root ownership is required for these configs to function is it not?

At the end of the day, you can put any restriction in this code you want. I think the point of my pull request is that it doesn't provide any real value. Making restrictions that are only based on principal is poor practice in my opinion. It only makes matters worse that the error message display doesn't accurately reflect the logic in the code.

"Ignoring %s because of bad file mode - must be 0644 or 0444"

rogerheflin:
Either the code is wrong, or the message of "bad file mode - must be 0644 or 0444" is. The code as written does not check for read permission on group or other.   So 0600 and 0400 also work.    The message  should be something like "execute is not allowed on anything, and you may not have write on group/other and you must not have any sticky bits set" as this is technically what the code currently implements.    If someone makes a mistake of the execution bit being set and this causes (for no good reason) logrotate to stop working on the newer versions.  It also does this silently (there is no message any any log file that is commonly viewed) and one has to turn on debugging to find out what is going on once you determine the files are not being rotated.   You reasoning for disallowing execute has no real point other than you wrote the code that way, and it may cause breakage for no useful reason.   Please attempt to put the smallest amount of restrictions on something that will meet the required security.

kdudka:
On Friday, January 19, 2018 5:54:39 PM CET thehomerepot wrote:
> In this situation, the attacker would need root access.

Sure.  If you have full root access already, you do not need to bother with 
logrotate configuration files.  If you have limited control over a daemon 
running as root, some outstanding permission bits here and there may help
you to construct the attack vector.

> If this is the stance, I don't understand why you don't restrict permissions
> to 600? Why does group or other need any access to these config files at
> all? Root ownership is required for these configs to function is it not?

I think that 0644 is very common mode for config files.  Many linux distros 
default to umask 022, which causes new files to be created with mode 0644.  
Insisting on using 0600 for logrotate config files would break a lot of setups 
for no real benefit.

> At the end of the day, you can put any restriction in this code you want. I
> think the point of my pull request is that it doesn't provide any real
> value.

It helped you to discover files with suspiciously set mode on your system.

> Making restrictions that are only based on principal is poor
> practice in my opinion. It only makes matters worse that the error message
> display doesn't accurately reflect the logic in the code.
> 
> "Ignoring %s because of bad file mode - must be 0644 or 0444"

True.  Would it help to say "should be" instead of "must be"?

kdudka:
On Friday, January 19, 2018 7:21:20 PM CET rogerheflin wrote:
> You reasoning for disallowing execute has no real point other than you wrote
> the code that way, and it may cause breakage for no useful reason.

Except that I did not write the code in question ;-)

thehomerepot:
> If you have limited control over a daemon
running as root, some outstanding permission bits here and there may help
you to construct the attack vector.

Logrotate already has the ability to execute code within those config files whether they have execute permissions or not. I don't think there is honestly a realistic scenario where having the execute bit set would cause or allow any harm.

> Insisting on using 0600 for logrotate config files would break a lot of setups
for no real benefit.

Which is the same argument I have for disallowing the execute bit for root. Interesting.

> It helped you to discover files with suspiciously set mode on your system.

No, it actually caused a file system to fill up after a migration from RHEL 6 to RHEL 7 because someone set an execute bit thinking it wouldn't matter (and it shouldn't). This was prior to v3.8 of logrotate when no permissions were checked for at all.

> True.  Would it help to say "should be" instead of "must be"?

No, at minimum if you're going to leave unnecessary restrictions in the code, please at least document them. You need to specify what the code is actually looking for and why the utilization of a config file is failing. Why leave an error message so vague?

kdudka:
On Sunday, January 21, 2018 3:04:25 AM CET thehomerepot wrote:
> > Insisting on using 0600 for logrotate config files would break a lot of
> > setups for no real benefit.
> 
> Which is the same argument I have for disallowing the execute bit for root.

Because you misunderstood and misquoted my reply.  My argument was that
files are created with 0644 _by default_ on many systems.  In your case,
the executable bit on your config file was set _by mistake_.

If your program complains about the default situation, it is only disturbing, 
whereas, if your program complains about a suspicious situation, most likely 
caused by mistake, it is useful.

> No, at minimum if you're going to leave unnecessary restrictions in the
> code, please at least document them. You need to specify what the code is
> actually looking for and why the utilization of a config file is failing.
> Why leave an error message so vague?

Fair enough.  What about printing a diagnostic message under the current 
conditions but reject the config file only if it is writable by group or 
others?  Does that work for you?

thehomerepot:
> Fair enough.  What about printing a diagnostic message under the current
conditions but reject the config file only if it is writable by group or
others?  Does that work for you?

That seems fair.

glensc:
don't reject config if it's group writable please.

i have `umask 0007` and if the files are version controlled (CVS) the files get 0660 permissions. which is rather secure value.

only makes sense to reject world writable configs.

rogerheflin:
I can see the position of group writable being a security issue.   About
the only way one could implement it would be to allow below a certain group
number, and that is probably not reasonable as that number cannot be
trusted to be consistent across distributions.   If say the group were
"user" that had write access and all users on the system have that group
then that would be a security issue.   Even if a user saved the file and
that caused logrotate to not work because of it having a bad owner then one
would have to hope that whoever fixes it replaces the file with a good one
and/or verifies that nothing bad is in the file and does not just change
the owner to be correct and let the other users mods run.

Our config files that a consistent across the environment are being
delivered by RPMs, or are being set by a script.  The RPM and/or scripts
pieces are what we put in version control.

kdudka:
On Sunday, January 21, 2018 10:49:18 PM CET Elan Ruusamäe wrote:
> don't reject config if it's group writable please.

@glensc Such config files have been rejected since logrotate-3.8.0.

> i have `umask 0007` and if the files are version controlled (CVS) the files
> get 0660 permissions. which is rather secure value.

So how come you have not hit the problem yet?

> only makes sense to reject world writable configs.

I think that security requirements highly depend on the target environment.  
Nobody has reported any real-world problem with group-writable config files 
rejected by logrotate yet.  I am not sure whether lowering the requirements 
proactively is desirable here...

PR #168, opened 2018-01-31
Move wtmp/btmp definitions to logrotate.d

dgadmin:
The definitions for wtmp and btmp, when included in the main config
file, make it very difficult to manage with CM tools
(e.g. Ansible). It would require complex logic to parse blocks within
the global file.

This patch moves them to logrotate.d directory, where they follow the
same logic as any other include from various packages, and can be easily
manipulated on per-line basis.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,7 +4,11 @@ All notable changes to this project will be documented in this file.
 
 ## [UNRELEASED]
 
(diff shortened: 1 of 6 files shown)


kdudka:
Looks good.  Thanks!

I would just update those comments such that ```examples/btmp``` does not speak about ```wtmp``` and vice versa.

Issue #38, opened 2016-06-04
allow several matches for one log file, most specific wins


CoolCold:
I'd be nice to allow (and not produce error message) in case of several matches for log. For example, i have /var/log/nginx/ dir, where every nginx log is going. Some specific logs, which are used for statistics collection (machine parsable format, for example), can be rotated everyday, and kept for 1 day only. But if i produce two entires, like:
```
"/var/log/nginx/server.com.access.reqsize.log" {
      daily
..
}
```
and
```
/var/log/nginx/*.log {
        weekly
...}
```
i got the error on stderr (but exit code is 0) about duplication.
Most specific file match (pattern length which matched?) should win and do not produce errors.

johnlinp:
Hi, I would like to take on this feature and implement it.

I will define the following functions:

1. `compareGlobPattern(pattern1, pattern2)`: it compares 2 glob patterns, and returns 1 if `pattern1` is more specific than `pattern2`, returns -1 if `pattern2` is more specific than `pattern1`, returns 0 if the comparison is not possible.
2. `compareGlobPatternGroup(patternGroup1, patternGroup2)`: it compares 2 glob pattern groups, and returns 1 if one of the pattern in `patternGroup1` is more specific than all of the patterns in `patternGroup2`, returns -1 vice versa, returns 0 if no pattern is more specific than all the patterns in the other pattern group.

For example, the following 2 configs:

```
/var/log/a.* /var/log/a.b.c.* {
   daily
...
}

/var/log/a.b.* /var/log/a.b.c.d.* {
   weekly
...
}
```

They makes 2 pattern groups:

```
patternGroup1: /var/log/a.*, /var/log/a.b.c.*
patternGroup2: /var/log/a.b.* /var/log/a.b.c.d.*
```

`patternGroup2` will win because `/var/log/a.b.c.d.*` is more specific than all the patterns in `patternGroup1`. Therefore, the log file `/var/log/a.b.c.d.e` will be rotated weekly.

Any comments would be appreciated.

kdudka:
I do not think we can (or want to) define strict order on these patterns.  How would you compare ```/var/log/a.*``` with ```/var/log/*.b```?

johnlinp:
No, I didn't have the intention to do so. In your example, I will treat it as duplicate configs, just like before.

johnlinp:
More specifically, I will define `compareGlobPattern(pattern1, pattern2)` as the following logic:

[...]

PR #173, opened 2018-02-07
config: ignore duplicated entries

johnlinp:
Closes: #38

### DIFF
--- a/config.c
+++ b/config.c
@@ -155,6 +155,8 @@ static int glob_errno = 0;
 
 static int readConfigFile(const char *configFile, struct logInfo *defConfig);
 static int globerr(const char *pathname, int theerr);
+static int globFiles(struct logInfo *log);
+static int handleOldDir(struct logInfo *log);
 
 static char *isolateLine(char **strt, char **buf, size_t length) {
 	char *endtag, *start, *tmp;
@@ -515,6 +517,9 @@ static void copyLogInfo(struct logInfo *to, struct logInfo *from)
     }
 	if (from->dateformat)
 		to->dateformat = strdup(from->dateformat);
+	if (from->configFilePath)
+		to->configFilePath = strdup(from->configFilePath);
+	to->configFileLineNum = from->configFileLineNum;
 }
 
 static void freeLogInfo(struct logInfo *log)
@@ -534,6 +539,7 @@ static void freeLogInfo(struct logInfo *log)
 	free(log->compress_ext);
 	free(log->compress_options_list);
 	free(log->dateformat);
+	free(log->configFilePath);
 }
 
 static struct logInfo *newLogInfo(struct logInfo *template)
@@ -684,6 +690,7 @@ int readAllConfigPaths(const char **paths)
 {
     int i, result = 0;
     const char **file;
+    struct logInfo *log = NULL;
     struct logInfo defConfig = {
 		.pattern = NULL,
 		.files = NULL,
@@ -748,6 +755,16 @@ int readAllConfigPaths(const char **paths)
     }
     free_2d_array(tabooPatterns, tabooCount);
     freeLogInfo(&defConfig);
+
+    for (log = logs.tqh_first; log != NULL; log = log->list.tqe_next) {
+        if (globFiles(log)) {
+            result = 1;
+        }
+        if (handleOldDir(log)) {
+            result = 1;
+        }
+    }
+
     return result;
 }
 
@@ -823,6 +840,172 @@ static char* parseGlobString(const char *configFile, int lineNum,
     return NULL;
 }
 
+static int findDuplicateFile(const char *filepath)
+{
+	int k;
+	struct logInfo *log;
+
+	for (log = logs.tqh_first; log != NULL; log = log->list.tqe_next) {
+		for (k = 0; k < log->numFiles; k++) {
+			if (!strcmp(log->files[k], filepath)) {
+				//message(MESS_ERROR, "%s:%d duplicate log entry for %s\n", configFile, lineNum, globResult.gl_pathv[glob_count]);
+				return 1;
+			}
+		}
+	}
+
+	return 0;
+}
+
+static int globFiles(struct logInfo *log)
+{
+	int rc;
+	const char **argv;
+	int argc, argNum;
+	int result = 0;
+	size_t glob_count;
+	struct stat sb;
+
+	if (poptParseArgvString(log->pattern, &argc, &argv)) {
+		message(MESS_ERROR, "%s:%d error parsing filename\n", log->configFilePath, log->configFileLineNum);
+		free(log->pattern);
+		log->pattern = NULL;
+		return 1;
+	} else if (argc < 1) {
+		message(MESS_ERROR, "%s:%d { expected after log file name(s)\n", log->configFilePath, log->configFileLineNum);
+		free(log->pattern);
+		log->pattern = NULL;
+		return 1;
+	}
+
+	log->files = NULL;
+	log->numFiles = 0;
+	for (argNum = 0; argNum < argc; argNum++) {
+		glob_t globResult;
+		rc = glob(argv[argNum], GLOB_NOCHECK
+#ifdef GLOB_TILDE
+												| GLOB_TILDE
+#endif
+											, globerr, &globResult);
+		if (rc == GLOB_ABORTED) {
+			if (log->flags & LOG_FLAG_MISSINGOK) {
+				continue;
+			}
+
+			message(MESS_ERROR, "%s:%d glob failed for %s: %s\n", log->configFilePath, log->configFileLineNum, argv[argNum], strerror(glob_errno));
+
+			globResult.gl_pathc = 0;
+		}
+
+		log->files = realloc(log->files, sizeof(*log->files) * (log->numFiles + globResult.gl_pathc));
+
+		for (glob_count = 0; glob_count < globResult.gl_pathc; glob_count++) {
+			/* if we glob directories we can get false matches */
+			if (!lstat(globResult.gl_pathv[glob_count], &sb) && S_ISDIR(sb.st_mode)) {
+				continue;
+			}
+
+			if (findDuplicateFile(globResult.gl_pathv[glob_count])) {
+				message(MESS_ERROR, "%s:%d duplicate log entry for %s\n", log->configFilePath, log->configFileLineNum, globResult.gl_pathv[glob_count]);
+				result = 1;
+				break;
+			}
+
+			log->files[log->numFiles] = strdup(globResult.gl_pathv[glob_count]);
+			log->numFiles++;
+		}
+		globfree(&globResult);
+	}
+
+	free(argv);
+
+	return result;
+}
+
+static int handleOldDir(struct logInfo *log)
+{
+	int i;
+	char *dirName;
+	struct stat sb1, sb2;
+
+	if (!log->oldDir) {
+		return 0;
+	}
+
+	for (i = 0; i < log->numFiles; i++) {
+		char *ld;
+		char *dirpath;
+
+		dirpath = strdup(log->files[i]);
+		dirName = dirname(dirpath);
+		if (stat(dirName, &sb2)) {
+			if (!(log->flags & LOG_FLAG_MISSINGOK)) {
+				message(MESS_ERROR,
+					"%s:%d error verifying log file "
+					"path %s: %s\n",
+					log->configFilePath, log->configFileLineNum, dirName, strerror(errno));
+				free(dirpath);
+				return 1;
+			} else {
+				message(MESS_DEBUG,
+					"%s:%d verifying log file "
+					"path failed %s: %s, log is probably missing, "
+					"but missingok is set, so this is not an error.\n",
+					log->configFilePath, log->configFileLineNum, dirName, strerror(errno));
+				free(dirpath);
+				continue;
+			}
+		}
+		ld = alloca(strlen(dirName) + strlen(log->oldDir) + 2);
+		sprintf(ld, "%s/%s", dirName, log->oldDir);
+		free(dirpath);
+
+		if (log->oldDir[0] != '/') {
+			dirName = ld;
+		}
+		else {
+			dirName = log->oldDir;
+		}
+
+		if (stat(dirName, &sb1)) {
+			if (errno == ENOENT && log->flags & LOG_FLAG_OLDDIRCREATE) {
+				int ret;
+				if (log->flags & LOG_FLAG_SU) {
+					if (switch_user(log->suUid, log->suGid) != 0) {
+						return 1;
+					}
+				}
+				ret = mkpath(dirName, log->olddirMode, log->olddirUid, log->olddirGid);
+				if (log->flags & LOG_FLAG_SU) {
+					if (switch_user_back() != 0) {
+						return 1;
+					}
+				}
+				if (ret) {
+					return 1;
+				}
+			}
+			else {
+				message(MESS_ERROR,
+					"%s:%d error verifying olddir "
+					"path %s: %s\n",
+					log->configFilePath, log->configFileLineNum, dirName, strerror(errno));
+				return 1;
+			}
+		}
+
+		if (sb1.st_dev != sb2.st_dev && !(log->flags & (LOG_FLAG_COPYTRUNCATE | LOG_FLAG_COPY | LOG_FLAG_TMPFILENAME))) {
+			message(MESS_ERROR,
+				"%s:%d olddir %s and log file %s "
+				"are on different devices\n",
+				log->configFilePath, log->configFileLineNum, log->oldDir, log->files[i]);
+			return 1;
+		}
+	}
+
+	return 0;
+}
+
 static int globerr(const char *pathname, int theerr)
 {
     (void) pathname;
@@ -858,24 +1041,17 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
     off_t length;
     int lineNum = 1;
     unsigned long long multiplier;
-    int i, k;
+    int i;
     char *scriptStart = NULL;
     char **scriptDest = NULL;
     struct logInfo *newlog = defConfig;
     char *start, *chptr;
-    char *dirName;
     struct passwd *pw = NULL;
-    int rc;
-    struct stat sb, sb2;
-    glob_t globResult;
-    const char **argv;
-    int argc, argNum;
+    struct stat sb;
 	int flags;
 	int state = STATE_DEFAULT;
     int logerror = 0;
-    struct logInfo *log;
 	static unsigned recursion_depth = 0U;
-	char *globerr_msg = NULL;
 	int in_config = 0;
 	int rv;
 	struct flock fd_lock = {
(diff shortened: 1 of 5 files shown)


johnlinp:
There are 2 commits here. The first one is a refactor and it shouldn't change any behavior of `logrotate`. The second one is the real fix, and it is very intuitive because it's based on the refactor.

Here is what I did:

- refactor commit: when parsing the glob pattern in the config file, I don't want to do the glob immediately. Instead, I just stored the glob pattern for later use. After all the `logInfo` is parsed, I do the glob for every `logInfo` sequentially. This is good because `missingok` can be handled right away instead of previous implementation:

```
/* We don't yet know whether this stanza has "missingok"
 * set, so store the error message for later. */
```

- real fix commit: I can handle `globoverride` in the glob loop, similar to how it handles `missingok`.

Note: a small catch on this is that I have to store the config file path and line number for each `logInfo` since I will have to print out the error message to show which log config contains errors.

Any comment would be appreciated.

kdudka:
Do I understand it correctly that, whenever a glob in logrotate's configuration matches a log file and sets something (or even sets nothing), then it can never be matched again to override the previous settings?

I would expect it to work the other way around (that later matches override the previous ones).  That would be more compatible with how logrotate handles multiple configuration files:

> Any number of config files may be given on the command line. Later config files may override the options given in earlier files, so the order in which the logrotate config files are listed is important.

... or with the principle that local definitions override global ones.

johnlinp:
Sorry, I don't understand. I thought `logrotate` doesn't allow 2 config files specify duplicate log files.

For example, I have a log at `/var/log/some-app.log` and 2 config files:

```
# a.conf
/var/log/some-app.log {
    rotate 1
    addextension .a
}
```

```
# b.conf
/var/log/some-app.log {
    rotate 1
    addextension .b
}
```

Then

```
# logrotate --force a.conf b.conf
```

will output:

```
error: b.conf:1 duplicate log entry for /var/log/some-app.log
```

and `/var/log/some-app.log` is rotated as `/var/log/some-app.log.1.a` according to `a.conf`.

It seems that `b.conf` didn't override the option. Did I miss something? Thanks.

kdudka:
On Wednesday, February 21, 2018 7:31:37 AM CET John Lin wrote:
> Sorry, I don't understand. I thought `logrotate` doesn't allow 2 config
> files specify duplicate log files.

Yes, that is the feature you are going to implement.  However, logrotate 
currently allows to specify multiple config files where _global_ options
from config files read earlier can be overridden by config files read later.

You want to implement a similar (optional) feature for per-log options.
What I am saying is that the handling of overrides should be consistent
with what is already implemented.  Hence, options for log files matched
later should override options for log files matched earlier and not vice 
versa.

Does it make sense now?

johnlinp:
If I understand correctly, you want the following behavior:

For the log files `/var/log/some-app-1.log` and `/var/log/some-app-2.log`,

```
# some-app.conf

/var/log/some-app-1.log {
    rotate 1
    addextension .a
}

/var/log/some-app-*.log {
    rotate 1
    addextension .b
    globoverride
}
```

You want both `/var/log/some-app-1.log` and `/var/log/some-app-2.log` to have extension of `.b` because the later log config should override the earlier one. However, I don't think it is desired since the later log config should be a "catch-all", as @wyohm described [in the comment](https://github.com/logrotate/logrotate/issues/38#issuecomment-359648259):

> So the above would rotate ...reqsize.log daily and everything else weekly.

In my current implementation, the above `logrotate` config will give `/var/log/some-app-1.log` the extension `.a` and give `/var/log/some-app-2.log` the exntension `.b`.

kdudka:
Then I would write those entries in the reverse order and still achieve the desired result.:
```
/var/log/some-app-*.log {
    rotate 1
    addextension .b
    globoverride
}

/var/log/some-app-1.log {
    rotate 1
    addextension .a
}
```

... 

Try the following two examples in a shell:
```
echo .b | tee /var/log/some-app-*.log
echo .a | tee /var/log/some-app-1.log
```

```
echo .a | tee /var/log/some-app-1.log
echo .b | tee /var/log/some-app-*.log
```

What will be the result?

johnlinp:
@kdudka I see your point. I will modify my patch to achieve this behavior. Thanks.

wyohm:
Was just about to write out my understanding of it, but it looks like it's agreed.

However, thinking about this made me consider how included files will behave. So if there are two conf files in `/etc/logrotate.d/` called `a.conf` and `b.conf`, will settings from `a.conf` still be in effect when `b.conf` is being processed? Or does the context get unwound between processing conf files?

kdudka:
@wyohm I do not think that file inclusion should change anything.  You should get the same result as if you inlined all the included config files.  Or am I missing something?

wyohm:
@kdudka That answers it, thanks. I'm not thinking that it will change behavior. Just that it highlights behavior which may not be desirable. For instance, if I want to understand the behavior of `/etc/logrotate.d/mariadb` I'd ideally want to only need to read `/etc/logrotate.d/mariadb` and `/etc/logrotate.conf`. But it's possible that `/etc/logrotate.d/httpd` (processed first because of alphabetical order) has set a global directive or a pattern with `globoverride` which may affect what's going on in `/etc/logrotate.d/mariadb`.

johnlinp:
Let's make the semantic meaning of `globoverride` clear:

In @wyohm's [original comment](https://github.com/logrotate/logrotate/issues/38#issuecomment-359648259), this directive means "this log config will do a glob search and take what are not handled in the earlier log configs".

In @kdudka's [comment](https://github.com/logrotate/logrotate/pull/173#issuecomment-368775281), this directive means "this log config will do a glob search and take all the glob results, but can be taken by the later log configs".

Do I understand this correctly?

kdudka:
@wyohm You have a similar problem even now without the ```globoverride``` option.  If any config file in ```/etc/logrotate.d``` sets a global option, it will have global effect, too.  I have quickly looked at files in ```/etc/logrotate.d``` and none of them sets global options.  But it is a matter of policy.  As far as I know logrotate itself does not guarantee that included files do not set global options.

wyohm:
@johnlinp Yes, that was the meaning of my original comment. Though I now prefer the behavior @kdudka has proposed. What I proposed may be simpler to implement and possibly understand conf files (log files can be processed as soon as they are matched), but what @kdudka proposes allows for more flexibility since multiple matches can add behavior.

kdudka:
@johnlinp Yes.  I think that the options should be applied to all matched log files.  And, if ```globoverride``` is used, the log files can be matched again and any options can be added, or previously used options can be overridden.

wyohm:
@kdudka Agreed, that issue (includes setting globals) currently exists. It's just that I only started thinking about it when considering the behavior of `globoverride`. It gives more opportunity for a problem. If anyone else thinks its an issue, I can write a feature request (e.g._unwind include context_). If not, I can live with it.

johnlinp:
Hmm.. I wasn't expect that later log config can **add** options to the previously matched logs. I think that **completely overriden** would be better. I think it's too complex for a system administrator to understand or debug since he/she has to accumulate all the effects in all matched log configs. If the behavior is **completely override**, it's easier to understand because the system administrator only needs to find out which the last matched log config is.

For example, for the 2 log files `/a-b.log`, `/a-c.log` and the following `logrotate` config:

```
daily
rotate 1

/a-*.log {
    globoverride
    rotate 2
}

/a-b.log {
    weekly
}
```

If the behavior is **adding** options, to figure out the options for `/a-b.log`, one should:

1. check global options: `daily` and `rotate 1`
1. see the first log config, find out it matches the pattern: change `rotate 1` to `rotate 2`
1. now it's `daily` and `rotate 2`
1. see the second log config, find out it matches the pattern again: change `daily` to `weekly`
1. it's finally `weekly` and `rotate 2`

If the behavior is **completely overriding** options, to figure out the options for `/a-b.log`, one can just:

1. check global options: `daily` and `rotate 1`
1. find out the last log config matches the pattern, which is the second one: `weekly`
1. put them together: `weekly` and `rotate 1`

In conclusion, I think the **add** would increase too much complexity and I suggest the **completely override** behavior. Thanks.

kdudka:
Let's keep it simple.  Currently it works like this:
- All logrotate config files are read in a specified order.
- Global options take effect on all log files.
- Local options take effect on log files matching the given glob.
- Local options override global ones.
- Later options override earlier ones.

We only need to (optionally) change one detail:
- A single log file can be matched by multiple globs.

Of course, you will be able to completely override ```daily``` by previously parsed ```weekly``` because they are mutually exclusive.  On the other hand, I do not think that ```nomail``` should invalidate already read ```weekly```.

By added complexity you mean complexity of the implementation?

As for complexity of the configuration interface, I see it exactly opposite.  Let's assume we have ```globoverride``` enabled globally.  If you want to quickly check that ```ifempty``` or ```notifempty``` is in effect on some logfile, you just grep ```/etc/logrotate*``` for ```ifempty``` and you will immediately see, which of them apply to the given log file.  Checking whether an unrelated ```nomail``` directive invalidated a previously accepted ```ifempty``` directive would be in my opinion more complex.

johnlinp:
By increasing complexity, I mean the complexity of the configuration interface. (Although I guess the complexity of the implementation would increase too)

Another question:

```
/*.log {
    daily
    globoverride
}

/a-*.log {
    weekly
}

/a-b-*.log {
    monthly
}
```

What option should `/a-b-c.log`have? Is it `weekly` because the second log config override the first one and the third one cannot override the second one? Or, is it `monthly` because the second one inherits `globoverride` from the first one so it can be overriden by the third one?

kdudka:
On Tuesday, February 27, 2018 4:49:58 PM CET John Lin wrote:
> By increasing complexity, I mean the complexity of the configuration
> interface. (Although I guess the complexity of the implementation would
> increase too)
> 
> Another question:
> 
> ```
> /*.log {
>     daily
>     globoverride
> }
> 
> /a-*.log {
>     weekly
> }
> 
> /a-b-*.log {
>     monthly
> }
> ```
> 
> What option should `/a-b-c.log`have? Is it `weekly` because the second log
> config override the first one and the third one cannot override the second
> one? Or, is it `monthly` because the second one inherits `globoverride`
> from the first one so it can be overriden by the third one?

Sounds pretty clear to me.  The options are applied as the config file is 
parsed:

1. daily and globoverride on all files matching /*.log
2. weekly on all files matching /a-*.log
3. monthly on all files matching /a-b-*.log

Consequently:

- /a-b-*.log files will rotate monthly
- /a-*.log files (not matching the above pattern) will rotate weekly
- /*.log files (not matching any of the above patterns) will rotate daily

johnlinp:
@kdudka I see. I will start working on it. Thanks!

johnlinp:
I would like to implement `globoverride` by the following algorithm in `readConfigFile()`:

- At reading glob pattern: print "duplicate log entry" error message only when the earlier log entry has no `globoverride` option.
- At reading "}": check if the log files overlaps any of the earlier log configs. If any log file in this log config `curr_log` overlaps an earlier log config `earlier_log`, then:
    - Create a new empty log config `new_log`
    - Copy `earlier_log` to `new_log` (with `copyLogInfo()`)
    - Use `curr_log` to override `new_log` (with to-be-implement `overrideLogInfo()`)
    - Take off the overlapped log files from `curr_log` and `earlier_log` and store them in `new_log`
    - Insert `new_log` into the global linked list `logs`

This algorithm make sure every log file will only appear at most once in the global linked list `logs`.

Any advice will be appreciated. Thanks.

kdudka:
Thanks for the analysis!  I will need to look at the current code closer myself.  Unfortunately I am too busy this week.  Please give me some time to process this.  Anyway, ```logrotate-3.14.0``` is going to be released next Friday.  So this feature will be targeted for ```logrotate-3.15.0```.

johnlinp:
By the way, I think the indentation in the code is a mess. The inconsistent of indentation makes it harder to develop. I would like to ask:

1. Can I make the indentation consistent in my commit?
1. If yes, do you prefer spaces or tabs?

Thank you.

Saruspete:
Hello,

There's way more lines starting with a tab than with a space, so I'd say tabs:

> grep -Pc '^[ ]+' *.h *.c
> log.h:1
> logrotate.h:34
> queue.h:109
> config.c:159
> log.c:2
> logrotate.c:422

> grep -Pc '^[\t]+' *.h *.c
> log.h:0
> logrotate.h:2
> queue.h:216
> config.c:1420
> log.c:56
> logrotate.c:1703

johnlinp:
@Saruspete Sorry for the late reply. Thanks for the statistics. I will use tab to reorganize the indent.

kdudka:
Sorry for being silent on this PR.  I have not found a sufficient block of time to have a deeper look at this.  As for (re)indentation and coding style changes beyond what is strictly necessary to implement this, those need to be discussed (and eventually pushed) independently of this PR.

The up to now practice was to resemble the indentation and coding style of the surrounding code.  It seems to cause problems in long term and it might now be the right time to unify the indentation and/or coding style to some extent.

However, such changes need to be discussed via appropriate channels, which a pull request named "config: ignore duplicated entries" is not.

johnlinp:
@kdudka Okay. I will create an independent issue for the indentation problem.

johnlinp:
The issue was created at https://github.com/logrotate/logrotate/issues/188.

Frank-Steiner:
Hi, I'd just like to ask about the state of this. We need to do major changes to our logrotate config due to the DSGVO and have to overwrite many of the default stuff that opensuse ships. It would make things very easy if we could just add confs to overwrite the handling of certain log files without  needing to change the shipped defaults. The current logrotate-master branch still seems to have the old behaviour.
cu,
Frank

johnlinp:
@Frank-Steiner Since the code indentation in current code base was kind of messy, I was waiting for #188 to close. If you need this feature, I can try to develop in current code base.

Frank-Steiner:
Yes, I understand that you'd like to know how to write the code before you start, so whatever you decide is fine with me. As soon as there's a patch I'll be happy to test it! Thanks!

kdudka:
Sorry for taking too long to review this PR.  I have to prioritize bug fixes over enhancements.  This PR changes fundamental parts of logrotate's code, so it needs to be reviewed carefully to make sure we are not introducing any bugs or compatibility issues in the future release of logrotate.

Saruspete:
Hello,

I'd really like to see this PR merged. 
The original ticket #38 is opened since 2016, and we're waiting its upstream merge to be back-ported to RH7 (customer case 02036032).

Could you see to prioritize it please ?
Thanks a lot !

johnlinp:
@Saruspete Sure. I am developing a new patch for the algorithm described in https://github.com/logrotate/logrotate/pull/173#issuecomment-369495685.

Saruspete:
Nice @johnlinp thanks for your work ! :)

Having reread the thread, here's what I understood : 
  - Original request is to have "most specific rule wins".
  - Current implementation is to skip the full block if any matched file is already handled in another rule. So the winner is the first configuration file to be read. Inclusion is done by alphanumerical sort
  - The globoverride would allow a first-timer rule to explicitely say "I can be overridden". If not, any rule coming afterwards (from sorting order) would generate an error (like currently)
  - If a more generic (wildcard) future rule is the glob with "globoverride", as previous, more specific ones are not flagged "globoverride", that would generate an error.
  - So naming of files is important, as it defines their priority in the processing list (and subsequent duplicates blocks to be skipped). However, packages just provide it with their name.

In the unix world, we are already used to prefix / suffix our configuration files to enforce alphanumeric ordering (and `export LC_ALL=C` to ensure consistency...). but we either prefix the whole listing with numbers (like in rc.d) or suffix the defaults we want with ".local" (like fail2ban). As packages just push a configuration file matching their program / package name, we're likely to use suffixes to keep things clean.

The suffix is useful so that packages updates don't generate wild ".rpmnew" files (that would be interpreted too), they just replace the packaged one, and reads the user suffixed afterwards.

With that in mind, the rules specified by user (the suffixed ones) are often discovered after the original (packaged) ones. 

If we'd want to say "I want this new rule to override the existing / package provided one", acting more like a "**OverrideExisting**".
If we'd want a "catch all not already matched" rule, we'd may just declare a wildcard glob with "**SkipExisting**" to say "If you find a duplicate file, skip me, keep using the old one".

Wouldn't that cover more real-world cases ?

kdudka:
I have not looked closely at the implementation yet but let's not over-complicate things.  The configuration rules need to be intuitive and as simple as possible.  What about the following rules?

- If `globoverride` (or whatever the configuration directive will be) is not used, everything continues to work as before.
- If `globoverride` is used for certain log files (either globally or by matched glob), it is possible to incrementally specify options for the affected log files by subsequent glob matches, possibly overriding any previously specified options.
- The order in which configuration files are read stays unchanged.
- The order in which glob results are processed stays unchanged (and by no means affect what is overridden and what not).

Am I missing anything?

johnlinp:
@kdudka I found that the algorithm described in [my previous comment](https://github.com/logrotate/logrotate/pull/173#issuecomment-369495685) cannot be easily done because the function `overrideLogInfo()` is impossible to be implemented with current `struct logInfo` definition. The reason `overrideLogInfo()` cannot be implemented is because `struct logInfo` contains lots of fields that cannot be determined if they are set by default value or set by config file.

For example, the field `threshold` defaults to be `1024 * 1024`. And I have 2 log info:

- `log_1`: has `threshold` of `512 * 1024`
- `log_2`: has `threshold` of `1024 * 1024`

When I want to override `log_1` with `log_2`, I have to determine whether `log_2.threshold` is a default value or set by user config file. If it's set by user config file, I should use it to override `log_1.threshold`. Otherwise (it's set by default) I should not use it.

However, current definition of `struct logInfo` does not contain such information about whether a field is set by default or set by user config file, so I am not able to implement the `overrideLogInfo()` function. May I add such information in `struct logInfo`? Will it change too much code for you?

kdudka:
I feel like I am repeating myself but why do we need all the complexity?  Cannot we just go through all config files and apply options incrementally as we read them?

kdudka:
@johnlinp I had finally time to look at the code.  I see what you are talking about -- your questions are pretty valid.  However, there is one more question that needs to be discussed before we proceed with the implementation:  What should we do with `sharedscripts`?

As it works now, each configuration section is associated with a set of globs and the actual set of files produced by expansion of the globs.  If a prerotate/postrotate script runs with `sharedscripts`, it gets a line-separated list of globs as its first argument.  However, if we implement the `globoverride` feature, there will not always be 1:1 mapping between the sets of globs and sets of files.

If we only update the set of files and keep the original set of globs while overriding a previous configuration, some files might be processed by multiple prerotate/postrotate scripts when `sharedscripts` is enabled.  That is certainly not what the user expects.  They should be able to override previously specified prerotate/postrotate script instead.  The problem is that the current script interface is simply not ready for that.

Would it make sense to pass line-separated list of files (instead of globs) to prerotate/postrotate scripts when both `sharedscripts` and `globoverride` are in effect?

johnlinp:
@kdudka I didn't notice there was a `sharedscripts` feature. Should we move the discussion for `sharedscripts` to the issue #38?

kdudka:
Good idea, moved: https://github.com/logrotate/logrotate/issues/38#issuecomment-432965529

johnlinp:
Accoding to the discussion in #38, I'm closing thie PR since I won't spend more time on it :(

PR #174, opened 2018-02-08
packaging: cleanups

ignatenkobrain:


kdudka:
Thank you for preparing the PR, Igor!  But I am afraid that upstream consumers expect the spec file to be more compatible than just with the currently supported releases of Fedora and RHEL.  You can see that the changes broke even our Travis CI job, which runs on a commonly used build environment (Ubuntu Precise).

glensc:
yes, there's no point of such cleanups (no breakage involved if cleanup not done). benefit is better coverage for older systems.

besides, downstream distributions have their own versions of .spec files anyway.

PR #178, opened 2018-03-01
fix build for freebsd

stha09:
Added some preprocessor checks to allow successful build on FreeBSD 11.1


kdudka:

PR #179, opened 2018-03-01
configure.ac: show support status for SELinux and ACL at the end

cgzones:


kdudka:

PR #180, opened 2018-03-04
fix cppcheck note about missing va_end and whitespace cleanup

cgzones:


kdudka:

PR #184, opened 2018-03-10
drop popt dependency in favor of getopt

cgzones:
**Not quite merge ready**

Removing libpopt dependency by importing from libpopt and re-implementing with getopt.

The command line parsing is re-implemented with getopt and should be roughly the same. Some differences are the error messages in cases of wrong parameters names or missing required arguments. Also `-h` is now an alias for `--help` and `-U` for `--usage`.

Two functions, `poptParseArgvString` and `poptDupArgv`, are needed for handling custom compress options. These two functions are copied and adapted for logrotate. I am not sure how handle the license matter correct for this case maybe someone could comment.

Are you willing to merge such kind of change? If yes I'll give it a deeper try and dig into the licensing question.

(The intention is to get rid of libpopt as it seems quite unmaintained.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -105,7 +105,7 @@ matrix:
 install:
   - sudo add-apt-repository ppa:dns/gnu -y
   - sudo apt-get -qq update
-  - sudo apt-get -qq install libpopt-dev libselinux1-dev libacl1-dev automake dash rpm
+  - sudo apt-get -qq install libselinux1-dev libacl1-dev automake dash rpm
 
 script:
   - $COMPILER --version
--- a/README.HPUX
+++ b/README.HPUX
@@ -13,28 +13,22 @@ also work on HP-UX 10.20):
     See http://devresource.hp.com/LPK/index.html for downloads.
     This library is needed to provide the ??? function.
  
-2.  Obtain, build, and install popt 1.4 (there doesn't seem to be a build at
-    the Porting Centre.)
-    See ftp://ftp.rpm.org/pub/rpm/dist/rpm-4.0.x/popt-1.6.4.tar.gz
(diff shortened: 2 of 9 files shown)


glensc:
`s/indention/intention/`

cgzones:
There are (to my knowledge) no issues which affect logrotate.
This pr was just a suggestion to drop a external dependency but its not a must.

kdudka:
If getopt was a drop-in replacement for popt, then the proposed change would make a lot o sense.  However, getting rid of an external dependency by taking over (a portion of) the source code of the external dependency does not sound like the best course of action.

kdudka:
Unless anybody comes with a strong reason for getting rid of the dependency on `popt` that would outweigh the troubles with maintaining the same code at multiple places, I am going to close this unresolved.

E5ten:
Now that the website that seems to be where popt is based as far as I can tell, rpm5.org is down and has been for ages, and it is clearly dead because the last commit according to where I can find the source (ubuntu's launchpad because like I said its upstream site is down and has been for ages) was in 2010, I feel like that should be sufficient reason to stop using popt. This also means maintaining the same code in multiple places is not an issue because there will be no upstream changes to keep watch for.

kdudka:
Nope.  The upstream just seems to move around:

https://github.com/rpm-software-management/popt/commits/master

E5ten:
Huh, well that's less dead than no commits since 2010, but that's still no releases since 2010 and no commits in over 2 years, which still feels pretty dead to me.

kdudka:
To me it feels that the software just works.  I see no open issues at logrotate that would be blocked by a bug or missing feature in popt.  I am not going to rework something that works (and risk that it would stop working) for no real benefit.

PR #186, opened 2018-03-18
small wording adjustments

cgzones:
also Debian puts package configurations in /etc/logrotate.d
GNU Public License -> GNU General Public License


kdudka:

PR #187, opened 2018-03-18
rename logrotate-default to logrotate.conf

cgzones:
this is the standard name used


kdudka:

PR #189, opened 2018-03-22
coding-style: unify indent style to tabs

### LINKED ISSUE #188: Unify code indentation
The indentation in the code is not consistent. Tabs and spaces are mixed. Some code blocks inside `for` or `if` are not indent properly. The problems increase the difficulty to understand the code.

I would like to submit a pull request to clean up the indentation. Any suggestions are welcome.

johnlinp:
Closes #188

### DIFF
--- a/config.c
+++ b/config.c
@@ -87,49 +87,49 @@ int asprintf(char **string_ptr, const char *format, ...)
 #if !defined(HAVE_STRNDUP)
 char *strndup(const char *s, size_t n)
 {
-       size_t nAvail;
-       char *p;
+	size_t nAvail;
+	char *p;
 
-       if(!s)
-               return NULL;
+	if(!s)
+		return NULL;
 
-       /* min() */
-       nAvail = strlen(s) + 1;
-       if ( (n + 1) < nAvail)
-               nAvail = n + 1;
+	/* min() */
+	nAvail = strlen(s) + 1;
+	if ( (n + 1) < nAvail)
+		nAvail = n + 1;
 
-       p = malloc(nAvail);
-       if (!p)
-               return NULL;
-       memcpy(p, s, nAvail);
-       p[nAvail - 1] = 0;
-       return p;
+	p = malloc(nAvail);
+	if (!p)
+		return NULL;
+	memcpy(p, s, nAvail);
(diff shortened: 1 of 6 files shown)


kdudka:
This seems to be more invasive than it needs to be.  I have just tried to blindly reindent the code with vim using the following setting:
```
set noet
set sw=4
set ts=8
```
... and ended up with smaller diffstat:
```
 config.c    | 2494 +++++++++++++++++++++++++++++------------------------------
 log.c       |  108 +--
 log.h       |    8 +-
 logrotate.c | 2012 +++++++++++++++++++++++------------------------
 logrotate.h |    6 +-
 5 files changed, 2314 insertions(+), 2314 deletions(-)
```

It could be probably further tweaked to avoid some changes that are not necessary.

johnlinp:
By setting `set sw=4`, the indent style would be 4 spaces. Do you prefer that?

kdudka:
> By setting set sw=4, the indent style would be 4 spaces.

That seems to be the most common style in logrotate sources.  Note that `sw` or `shiftwidth` on its own does not tell vim whether to use tabs or spaces.  That is what `(no)et` or `(no)expandtab` is for.

> Do you prefer that?

I prefer to make as small changes as possible while ending up with a uniformly looking coding style that is easy to maintain in long term.

Anyway, this is not about my personal preference.  I expect a wider discussion with the community before we make the final decision about how (or whether) to  reformat the source code of logrotate.  The indentation has been messy for quite a long time.  I do not see an immediate need to take an action on this.  Let's keep enough time for the discussion first!

glensc:
i totally agree with @kdudka that should choose style to follow most of the current style `vim: ts=4 sw=4 et`, even my personal preference is `vim: ts=4 sw=4 noet`.

i think the current mess is because vim's defaults: `vim: ts=8 sw=8 noet` and people tend to have `vim: ts=8 sw=4` somewhy.

ps: i think vim/emacs nodeline should be added to source files, so the editors can pick up the settings. and perhaps [.editorconfig](http://editorconfig.org/) to have more editor/ide agnostic settings.

kdudka:
This PR no longer applies.  Please comment on #212 and #213 instead.

PR #192, opened 2018-04-04
Warn user if both size and the time interval options are used

Burgos:
The documentation says (in the section for (max,min)size) that
the size option is mutually exclusive with the time interval option
and it says:

```
The related size option
is similar except that it is mutually exclusive with the  time
interval  options, *and it causes log files to be rotated with-
out regard for the last rotation time.*
```

(emphasis mine).

The emphasised bit of the sentence is not necessarily true - it is
true _if the size option is specified after the time option_. If
the time option is specified after the size option, the logrotate
will rotate the files without regard for the size. In other words -
the last option specified in the config file wins.

This behaviour is dangerous, because the subtle change in the order
of the configuration options changes the behaviour in the unexpected
way.

We can't treat this behaviour as an error, since the chances are that
there are many logrotate config files deployed in the way that it works
to the satisfactory degree, so that would be a breaking change. This
patch makes logrotate to issue a warning if logrotate run in verbose mode.


kdudka:
Thank you for pointing it out!  I agree with your reasoning but I am not happy with the actual implementation.  The newly introduced flags `time_criterium` and `size_criterium` are not handled per log, so the warning would trigger even for the following case, which is perfectly valid:

```
/var/log/a {
    size 2
}

/var/log/b {
    weekly
}
```

I would probably avoid using such flag at all and just warn immediately when a previously specified directive is being overridden.  We should probably also update the man page to say that directive specified the last time takes effect.

Burgos:
@kdudka thank you for pointing that out, that's a bug in the implementation, so I'm not happy with it as well. I'll follow your advice, as it sounds reasonable and I'll update the man pages as well over the weekend. I'll let you know when I push the updated version.

kdudka:
Sounds good.  Thank you for working on it and sorry for the embarrassing delay with my review!

Burgos:
@kdudka I've looked again into this, but I don't think there's a nice way to do this without flags. The problem is that the default criteria is set to `ROT_SIZE`, so I can't use check against it, to check if it's set _explicitly_ in the config file. However, I changed the code to actually reset the flags and display warning, once per definition (with the test case added to ensure this). I hope this is better, if you still don't like it, I'll look further for alternatives.

I've also updated the man page.

Burgos:
Ah the tests are failing with `dash`, will address.

Burgos:
Fixed the issue with the failing tests in `dash`.

kdudka:
Thank you for moving it forward!  I have slightly tweaked the check to be more generic (now it warns if any criterium is overridden) and the error message more specific.  Could you please have a look at the upstream `size_weekly` branch?

https://github.com/logrotate/logrotate/compare/size_weekly

I will later rename `test-0102` (likely to `test-0080`) but let's wait till #193 is merged.

Burgos:
@kdudka the changes looks good to me, from my side it's all good to merge that branch! Thank you for the improvement of the original patch!

PR #193, opened 2018-04-12
When sharedscript is not enabled, pass rotated log file name as a second argument to postrotate script.

dmccombs:
I've frequently wanted to know what the final rotated file name is of a log file in the `postrotate` script, especially when using a `dateext` with seconds that I can't easily predict. Rather than using workarounds with `find` to locate the correct log, this passes it in as the source of truth.

This second argument is `NULL` for any scripts other than `postrotate`, and is `NULL` if `sharedscripts` is enabled, since it doesn't make sense to only have the filename of the last log in that case.

A test checking for expected output for `prerotate` and `postrotate` included.


glensc:
some broken merge on changelog in this PR. perhaps rebase against origin/master.

dmccombs:
@glensc fixed the bad merge of `Changelog.md`.

logrotate.c
```diff
@@ -2145,8 +2145,8 @@ static int rotateLogSet(struct logInfo *log, int force)
 		message(MESS_DEBUG, "not running postrotate script, "
 			"since no logs were rotated\n");
 	    } else {
-		message(MESS_DEBUG, "running postrotate script\n");
-		if (runScript(log, log->flags & LOG_FLAG_SHAREDSCRIPTS ? log->pattern : log->files[j], log->post)) {
+		// It only makes sense to pass in a final rotated filename if scripts are not shared
```
kdudka: Please use `/* ... */` comments instead of `//` comments, which are used nowhere else in logrotate code.

test/Makefile.am
```diff
@@ -77,7 +77,8 @@ TEST_CASES = \
 	test-0076.sh \
 	test-0077.sh \
 	test-0100.sh \
-	test-0101.sh
+	test-0101.sh \
```
kdudka: To be honest, I do not know why the list of tests is sparse ... but please make the new test-cases `test-0078` and `test-0079`.

logrotate.c
```diff
@@ -2146,7 +2146,8 @@ static int rotateLogSet(struct logInfo *log, int force)
 			"since no logs were rotated\n");
 	    } else {
 		message(MESS_DEBUG, "running postrotate script\n");
-		if (runScript(log, log->flags & LOG_FLAG_SHAREDSCRIPTS ? log->pattern : log->files[j], log->post)) {
+		// It only makes sense to pass in a final rotated filename if scripts are not shared
```
kdudka: Please use `/* ... */` comments instead of `//` comments, which are used nowhere else in logrotate code.

dmccombs:
Updated with tests renumbered and comment style switched.

dmccombs:
@kdudka that squashed `rotated-log-postrotate-arg` branch works as expected for me.

kdudka:
Pushing now.  Thanks for the contribution!

dmccombs:
@kdudka I'm hoping to nudge the maintainers of the distro I use to update their logrotate version to include this change, is there any timeline on when a new release version will be tagged that includes this?

PR #195, opened 2018-04-27
copyTruncate: make 'copytruncate' work with 'rotate 0'

### LINKED ISSUE #191: "rotate 0" seems not to work with "copytruncate"
Hi, everyone! After some extensive testing, mostly on `3.9.2`, it seems that there's no way to truncate a file without creating a copy.

I heard that using `rotate 0` with `copytruncate` was likely supposed to work, but after no luck with that, I was pointed here thanks to help from @kdudka, who suggested I create an issue for others to look at (it didn't seem to be working on the latest version, as well).

I apologize if this is the normal, expected behavior or if I'm missing something simple, but I tried my best to use as many different configurations as possible to see if any would work. 🤔 

I appreciate your time, and I appreciate this awesome tool we use for other scenarios hundreds of times every week!

kdudka:
Closes #191

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -1047,6 +1047,7 @@ static int sparse_copy(int src_fd, int dest_fd, struct stat *sb,
 static int copyTruncate(char *currLog, char *saveLog, struct stat *sb,
 			int flags)
 {
+    int rc = 1;
     int fdcurr = -1, fdsave = -1;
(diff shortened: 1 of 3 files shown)


TheKLARKEN:
This looks awesome! I'm very excited about this! 🎊

PR #200, opened 2018-05-17
config / logrotate: fix couple printf format specifiers

kerolasa:
All of these are are warnings similar to one below.

config.c:369:59: warning: format ‘%d’ expects argument of type ‘int’, but
argument 4 has type ‘uid_t’ {aka ‘unsigned int’} [-Wformat=]

### DIFF
--- a/config.c
+++ b/config.c
@@ -366,7 +366,7 @@ static int do_mkdir(const char *path, mode_t mode, uid_t uid, gid_t gid) {
 			return -1;
 		}
 		if (chown(path, uid, gid) != 0) {
-			message(MESS_ERROR, "error setting owner of %s to uid %d and gid %d: %s\n",
+			message(MESS_ERROR, "error setting owner of %s to uid %u and gid %u: %s\n",
 				path, uid, gid, strerror(errno));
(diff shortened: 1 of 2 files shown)


kdudka:
I do not think that we can safely assume that `uid_t` is `unsigned int` on all supported platforms.  If we want to stay platform independent, we need to cast them explicitly to `unsigned int` before passing to printf.

kerolasa:
I see. So the most porbably correct way to print uid_t is to type cast it to signed int64_t before printing. Whether such overkill is better than seeing couple warnings on about type mismatches is questionable. In sake of completeness I migth update the pull to do that, unless adviced not to waste time with this.

kdudka:
I was about to say "let's not complicate things with `int*_t` types" but then I saw your commit 0c94a219 where you already used `intmax_t` in some debug messages.  Given the fact that nobody has complained about it, upstream consumers must have been able to cope with that.

Anyway, I think that casting the `uid_t` and `gid_t` values to `unsigned` for printing will work just fine to kill those warnings in a portable and well-defined way.

PR #213, opened 2018-08-10
style: reindent source code using spaces only

kdudka:
```
vim: et sw=4 ts=4
```
Pros:
- easier to view source code in a consistent way (as long as monospace font is used)
- people find it easier to contribute to logrotate when this coding style is used
- a trivial Travis CI hook can be installed to make sure that no tabs will appear in the source code again

Cons:
- more changes are needed to introduce this coding style:
```
 config.c    | 3062 +++++++++++++++++++++++------------------------
 log.c       |  112 +-
 log.h       |   14 +-
 logrotate.c | 3814 +++++++++++++++++++++++++++++------------------------------
 logrotate.h |   42 +-
 5 files changed, 3522 insertions(+), 3522 deletions(-)
```
- consequently, it is more difficult to backport patches to previous versions of logrotate

PR #226, opened 2018-09-12
preserve fractional part of timestamps when compressing

Villemoes:
When the system clock is carefully steered by NTP (in a way that it
never jumps), one can rely on no log file containing a timestamp newer
than the mtime of the file. However, with subsecond-precision
timestamps, that property is violated when using utime() to copy the
timestamps from the original log file to the compressed version.

utimensat() should be available on most modern systems (given that it
was standardized 10 years ago), but just in case, keep a fallback to
utime().


kdudka:
Looks good.  I will get back to this once #213 is merged (this Friday).

kdudka:
Merging now.  Thanks for the contribution!

PR #228, opened 2018-09-14
add some error checking

cgzones:


kdudka:
Thanks for the fixes!

What are the weird numbers in commit messages?

I am afraid they have no value for anybody but you.  I propose to use the following format instead:
```
Error: NEGATIVE_RETURNS:
config.c:597: negative_return_fn: Function "open(".", 0)" returns a negative number.
config.c:597: assign: Assigning: "here" = "open(".", 0)".
config.c:647: negative_returns: "here" is passed to a parameter that cannot be negative.
#  645|               qsort(namelist, files_count, sizeof(char *), compar);
#  646|           } else {
#  647|->             close(here);
#  648|               return 0;
#  649|           }
```
```
Error: CHECKED_RETURN:
logrotate.c:2353: check_return: Calling "stat(stateFilename, &sb)" without checking return value. This library function may fail and return an error code. [Note: The source code implementation of the function has been overridden by a builtin model.]
# 2351|   
# 2352|       close(fdcurr);
# 2353|->     stat(stateFilename, &sb);
# 2354|   
# 2355|       fdsave = createOutputFile(tmpFilename, O_RDWR | O_CREAT | O_TRUNC, &sb, prev_acl, 0);
```

cgzones:
updated commit messages

PR #235, opened 2019-01-16
Mitigation of a race condition in createOutputFile()

whotwagner:
If the "create"-option is set, there is a time window between rotating the logfile
and creating/touching a new empty file. If a user is in control of the parent
directory, he/she could replace it during this time window with a symbolic link to
any directory. If logrotate was executed by root, the new file will be created first
with permissions of root and then the ownerships will be set to the user. After the
file was created, the user might be able to write any content to this file.


kdudka:
First off, thank you for helping us to improve logrotate!  But I do not think that your change could reliably mitigate the attack vector you were describing for multiple reasons:

1. The `createOutputFile()` function itself is already resistant to the symlink-based attacks because it operates on a file descriptor.  So if the file with the originally opened name is replaced after the call of `open()`, the `createOutputFile()` function continues to use the **opened file** and it is not affected by the replacement.  The problem you were describing happens between the `rename()` operations, as I understand it.  I believe that it can happen while using `nocreate`, too.  So a reliable fix simply cannot be implemented in `createOutputFile()`.

2. You call `lstat()` on the **name** of the parent directory.  So if a non-privileged user has write access to any (direct or indirect) ancestor of the parent directory, which you patch does not check for, they can replace the directory by anything else after your call of `lstat()`.

3. The symlink does not need to be created at the parent directory directly.  You would need to check also all indirect parents up to the root.  And even then the code would be prone to race conditions unless you rewrite logrotate to operate on open file descriptors only.  This would basically mean to start using `openat()`, `renameat()`, etc. everywhere in the code.

The change also introduces a memory leak but it is not important at this step of the review.

kdudka:
Yes, it is not easy to make 20+ years old software immediately perfectly secure.  On the other hand, we should at least improve the documentation.  It currently roughly says what the `su` directive does but the documentation fails to explain why administrators should proactively use the `su` directive in certain cases.

kdudka:
I have submitted pull request #236 to improve the documentation.

Issue #238, opened 2019-02-06
Exit code 1, no errors in -d 


goldencut:
Was updating Logrotate configs with Ansible on multiple servers and noticed exit codes >0 on some when task ran 'logrotate -f /etc/logrotate.conf':
`TASK [logrotate : Reload conf] ****************************************************************************************
fatal: [myserver.com]: FAILED! => {"changed": true, "cmd": ["logrotate", "-f", "/etc/logrotate.conf"], "delta": "0:00:06.372079", "end": "2019-02-06 15:55:18.817965", "msg": "non-zero return code", "rc": 1, "start": "2019-02-06 15:55:12.445886", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}`

When I run on server: 
`root@myserver.com:~# logrotate -f /etc/logrotate.conf || echo "notOK"`
`notOK`
`root@myserver.com:~# logrotate -f /etc/logrotate.conf ; echo $?`
`1`

But I get no errors when I run:
`logrotate -d /etc/logrotate.conf 2> out.txt && grep -i error out.txt`

Is this by design somehow, am I missing something?

BR,
G.

kdudka:
`logrotate -d` does nothing, only debug messages are printed.  You need to use `logrotate -v` to get verbose output of logrotate that actually does something.

goldencut:
Thanks. 
I ran: 
`logrotate -v -f /etc/logrotate.conf 2> out.txt; grep -i error out.txt` 
No errors.
On what conditions does logrotate set exit code to 1? I assumed 'already exists, skipping rotation' shouldn't cause that.

kdudka:
It is not clear which version of logrotate you are using, neither how logrotate's configuration looks like.  Nevertheless, the general answer is yes, this may happen.  See the source code:

https://github.com/logrotate/logrotate/blob/7fae5f2bba0f20893394630d42ea196a4ff6df21/logrotate.c#L1833

goldencut:
I got my answers, IIUC there were improvements made to code... who closes the ticket? 

PS. very minimal logrotate conf, v.3.8.7 (latest for Debian 8)

kdudka:
Glad to see it resolved, closing...

Note that `logrotate-3.8.7` was released more than 5 years ago.  Unless Debian provides support for the very old release, you will need to upgrade it to a more recent version (`logrotate-3.15.0` was released in December 2018).

PR #239, opened 2019-02-08
improved error messages

kplimack:
# What
Improving error messaging.
* This PR adds the "error" prefix to messages that make the exit code non-zero, making the output simple to `grep` "error".  
* Output the total number of errors detected before exiting

# Why?
When certain pieces of code change the exit-code to 1 and don't say the word error in their output, it makes it challenging to determine why logrotate is failing its checks without diving into the code.

## Example
From the output of `logrotate -d`, it looks like there are no problems, it just exits 1.  So you comb through the output looking for a message that indicates the reason, but nothing looks like a failure. You see a message like
>destination /var/log/something/something already exists, skipping rotation

This is an error and is raising the exit code, but it is unclear that this is the behavior that is causing the failure

Related to https://github.com/logrotate/logrotate/issues/238

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -1530,7 +1530,7 @@ static int prerotateSingleLog(struct logInfo *log, int logNum,
         while ((*dext != '\0') && (!hasErrors)) {
             /* Will there be a space for a char and '\0'? */
             if (j >= (sizeof(dext_pattern) - 1)) {
-                message(MESS_ERROR, "error: Date format %s is too long\n",
+                message(MESS_ERROR, "Date format %s is too long\n",
                         log->dateformat);
                 hasErrors = 1;
                 break;
@@ -1552,7 +1552,7 @@ static int prerotateSingleLog(struct logInfo *log, int logNum,
                                 sizeof(dext_pattern) - strlen(dext_pattern) - 1);
                         j += 10;
                         if (j >= (sizeof(dext_pattern) - 1)) {
-                            message(MESS_ERROR, "error: Date format %s is too long\n",
+                            message(MESS_ERROR, "Date format %s is too long\n",
                                     log->dateformat);
                             hasErrors = 1;
                             break;
@@ -1567,7 +1567,7 @@ static int prerotateSingleLog(struct logInfo *log, int logNum,
                                 sizeof(dext_pattern) - strlen(dext_pattern) - 1);
                         j += 50;
                         if (j >= (sizeof(dext_pattern) - 1)) {
-                            message(MESS_ERROR, "error: Date format %s is too long\n",
+                            message(MESS_ERROR, "Date format %s is too long\n",
                                     log->dateformat);
                             hasErrors = 1;
                             break;
@@ -1831,7 +1831,7 @@ static int prerotateSingleLog(struct logInfo *log, int logNum,
         }
         if (!stat(destFile, &fst_buf)) {
             message(MESS_ERROR,
-                    "error: destination %s already exists, skipping rotation\n",
+                    "destination %s already exists, skipping rotation\n",
                     rotNames->firstRotated);
             hasErrors = 1;
         }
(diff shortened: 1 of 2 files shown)


kdudka:
Thank you for raising this topic!  I think that vast majority of the `message(MESS_ERROR, ...)` calls should result in non-zero exit code.  The few exceptions could be fixed to either return non-zero code, or by replacing `MESS_ERROR` by a different level.  We could probably introduce `MESS_WARNING` to print warnings.  Then we can easily modify the `message()` function to always print the `error: ` prefix for `MESS_ERROR` and `warning: ` for `MESS_WARNING`, instead of duplicating the prefixes all over the code.  Does it make sense?

andy-v-h:
It does make sense. New push coming shortly.

logrotate.c
```diff
@@ -2811,6 +2811,9 @@ int main(int argc, const char **argv)
     if (!debug)
         rc |= writeState(stateFile);
 
+    if (debug && rc)
+        message(MESS_ERROR, "logrotate has had %d errors",  rc);
```
kdudka: This output is misleading. `rc` is a return code, not count of errors.

kdudka:
This pull request becomes messy as there are a lot of commits reverting each other.  The resulting diff just adds the debug/info/verbose/error/fatal/unknown prefix, as I understand it.  Note that the debug prefix does not work well for example with `rotateLogSet()`, where it prints something like this:
```
debug: reading config file lr.conf
debug: Reading state from file: lr.stat
debug: Allocating hash table for state file, size 64 entries
debug: Creating new state
debug: 
Handling 1 logs
debug: 
rotating pattern: /tmp/xxx  debug: 1048576 bytes debug: (1 rotations)
debug: empty log files are rotated, debug: old logs are removed
debug: considering log /tmp/xxx
debug:   Now: 2019-11-04 13:26
debug:   Last rotated at 2019-11-04 13:00
debug:   log does not need rotating (log size is below the 'size' threshold)
debug: set default create context to unconfined_u:object_r:user_home_t:s0
```

andy-v-h:
@kdudka I apologize for the messy commit history, I’ve been working in Go mostly these days, and didn’t know how to best build and run test locally. I ended up having a couple of script locally. 

 I hadn’t considered how it would pollute the output. I hadn’t been concerned with that since I’m used to writing syslog formatted messages so the log level being on each line feels normal but I understand the need to minimize behavior changes. Perhaps an additional flag for whether or not to pretend with `debug: ` then entire blocks would look more like 

```
debug: reading config file lr.conf
       Reading state from file: lr.stat
       Allocating hash table for state file, size 64 entries
       Creating new state
       
Handling 1 logs
       
rotating pattern: /tmp/xxx  debug: 1048576 bytes debug: (1 rotations)
       empty log files are rotated, debug: old logs are removed
      considering log /tmp/xxx
         Now: 2019-11-04 13:26
         Last rotated at 2019-11-04 13:00
         log does not need rotating (log size is below the 'size' threshold)
       set default create context to unconfined_u:object_r:user_home_t:s0 
```
That way each time a series of debug level logs comes through, it is only pretended once?

kdudka:
Can't we just use the prefix for error messages only?  Those affect logrotate's exit code after all...

paumic:
Not sure if this is related place to post, but it seems that `destination /my/rotated/file/new-name already exists, skipping rotation` and `log does not need rotating (log size is below the 'size' threshold)` messages both goes to the stderr. 

First type will trigger RC 1, while second type will return RC 0. Perhaps it would be possible to either:

Place output to stdout which does not trigger RC 0, and/or separate error messages in stderr to determine which ones does trigger RC 1.

cgzones:
> Not sure if this is related place to post, but it seems that `destination /my/rotated/file/new-name already exists, skipping rotation` and `log does not need rotating (log size is below the 'size' threshold)` messages both goes to the stderr.
> 
> First type will trigger RC 1, while second type will return RC 0. Perhaps it would be possible to either:
> 
> Place output to stdout which does not trigger RC 0, and/or separate error messages in stderr to determine which ones does trigger RC 1.

Currently all messages are printed to `stderr`
https://github.com/logrotate/logrotate/blob/5b4518df05253c6956304088cb9a37dd4df04dee/log.c#L60-L64
but the message ```log does not need rotating (log size is below the 'size' threshold)``` is only printed in debug mode
https://github.com/logrotate/logrotate/blob/5b4518df05253c6956304088cb9a37dd4df04dee/logrotate.c#L1427-L1430

PR #240, opened 2019-02-09
config: add variable initialization [-Wconditional-uninitialized]

kerolasa:
Noticed when compiling with clang.

config.c:314:17: warning: variable 'm' may be uninitialized when used here
config.c:287:19: note: initialize the variable 'm' to silence this warning


kdudka:
This is a false positive.  There is no feasible code path that would use an uninitialized value of `m` (unless you have a broken implementation of `sscanf()`, in which case anything may happen).  Nevertheless, the patch is harmless, so I am merging it...

PR #258, opened 2019-08-26
.travis.yml: use Ubuntu Trusty (a.k.a. 14.04)

kdudka:
... because Ubuntu Precise (a.k.a. 12.04) is EOL


kdudka:
Thanks for review!

PR #264, opened 2019-09-09
man page: style and formatting changes

cgzones:
- Have two space character between sentences or move the beginning sentence to the next line (preferred)
- Section numbers of manuals is set in roman type
- Use the name of a unit in text
- Print an escape character with '\e', not '\\'
- Reduce space between words to one space character
- Use '\(en' for a dash meaning a range
- Use a one font macro when there is only one argument
- Use for an ellipsis use the string "\&.\|.\|.\|\&"
- "left" in a table is not valid, use just 'l'

Based on patch from Bjarni I. Gislason <bjarniig@rhi.hi.is> over at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=915301

### DIFF
--- a/logrotate.8.in
+++ b/logrotate.8.in
@@ -28,7 +28,7 @@ a log more than once in one day unless the criterion for that log is
 based on the log's size and \fBlogrotate\fR is being run more than once
 each day, or unless the \fB\-f\fR or \fB\-\-force\fR option is used.
 .P
-Any number of config files may be given on the command line. Later config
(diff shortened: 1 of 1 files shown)


kdudka:
@cgzones Has this already been reviewed by someone?  I do not feel competent to review it myself...

cgzones:
No, I am not an expert in writing man page either, but I pinged the original path author.

kdudka:
Thanks!  Merging...

PR #265, opened 2019-09-10
man page: mention cfg files must not be group-writable nor world-writable

cgzones:
Originated from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=691407


kdudka:
Thanks!  Merging...

PR #268, opened 2019-09-25
Date/Time format and debug messages improvements

szaydel:
Hi! While doing a little bit of debugging with `logrotate` I noticed that timestamps did not have proper format, at least values did not always take two places, and pad with zero where a value is less than 10. I also observed messages being somewhat less than well-formed, so decided to do a little bit of cleanup. Hopefully this will be acceptable. It certainly looks better now.

Some examples of resulting changes:
```
<   log does not need rotating (log has been rotated at 2019-9-12 4:0, that is not month ago yet)
---
>   log does not need rotating (log has been rotated at 2019-09-12 04:00, which is less than a month ago)

<   log does not need rotating (log has been rotated at 2019-9-22 4:2, that is not week ago yet)
---
>   log does not need rotating (log has been rotated at 2019-09-22 04:02, which is less than a week ago)
```
Let me know if this is acceptable, or if you would like something else done, etc. Feel free to close if for whatever reason this is not acceptable.

Thanks, Sam.

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -1266,8 +1266,8 @@ static int findNeedRotating(struct logInfo *log, int logNum, int force)
                             && now.tm_wday == log->weekday);
                 if (!state->doRotate) {
                     message(MESS_DEBUG, "  log does not need rotating "
-                            "(log has been rotated at %d-%d-%d %d:%d, "
-                            "that is not week ago yet)\n", 1900 + state->lastRotated.tm_year,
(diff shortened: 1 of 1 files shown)


kdudka:
Looks good to me.  Thanks for the contribution!  Merging now...

szaydel:
Thanks a lot, I appreciate it @kdudka!
Cheers,
Sam.

PR #272, opened 2019-10-07
some fixes

cgzones:
- main: remove unneeded cast
- main: contify logfile argument
- hashIndex: whitelist intentional unsigned integer overflow
- prerotateSingleLog: fix memory leak
- prerotateSingleLog: fix heap-buffer-overflow
- removeLogInfo: fix memory leak
- copyLogInfo: copy queue information to avoid SEGV

see single commits for more info


kdudka:

PR #273, opened 2019-10-07
make lgtm.com happy

cgzones:
- logrotate.c: convert localtime to localtime_r
- logrotate.c: drop ancient FIXME

You might also consider enabling [lgtm.com](https://lgtm.com/projects/g/logrotate/logrotate/ci/) for pull requests

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -384,11 +384,13 @@ static void restoreSecCtx(void **pPrevCtx)
 
 static struct logState *newState(const char *fn)
 {
-    struct tm now = *localtime(&nowSecs);
+    struct tm now;
     struct logState *new;
     time_t lr_time;
 
     message(MESS_DEBUG, "Creating new state\n");
+
+    localtime_r(&nowSecs, &now);
 
     if ((new = malloc(sizeof(*new))) == NULL)
         return NULL;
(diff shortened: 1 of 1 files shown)


logrotate.c
```diff
@@ -1454,6 +1458,7 @@ static int prerotateSingleLog(struct logInfo *log, int logNum,
     if (log->compress_ext && (log->flags & LOG_FLAG_COMPRESS))
         compext = log->compress_ext;
 
+    localtime_r(&nowSecs, &now);
```
kdudka: Could we optimize out the `now` variable and write to `state->lastRotated` directly?

logrotate.c
```diff
@@ -1286,7 +1290,7 @@ static int findNeedRotating(struct logInfo *log, int logNum, int force)
                 }
                 break;
             case ROT_DAYS:
-                /* FIXME: only days=1 is implemented!! */
+                /* TODO: only days=1 is implemented!! */
```
kdudka: I fail to see how TODO fits better here than FIXME.  The comment is there at least since 1996 and I have not seen any requests by logrotate's users to implement it.  I think it should be safe to just remove the comment.

logrotate.c
cgzones: I do not think so; `now` is later used in the function https://github.com/logrotate/logrotate/blob/5e7cecb677e4a35aa77423d7aff9304061327013/logrotate.c#L1510

kdudka: Indeed.  My bad.

kdudka:
Thanks!  Merging...

PR #274, opened 2019-10-07
logrotate.service: enable systemd hardening option ProtectKernelTunables

cgzones:
systemd version 235 is now two years old
see https://github.com/systemd/systemd/commit/3a0bf6d6aa08a267aefd5d5620fb6bee2556affb


kdudka:
Thanks!  Merging...

PR #277, opened 2019-10-09
Fix #276: error early if su lacks a user or group

### LINKED ISSUE #276: Handle missing group in su better
Currently, if you use `su` and only give it a user (not a group too), it results in an error like this:

```
error: error switching euid to 2048 and egid to -1: Invalid argument
```

It's not at all obvious what the problem is here. Instead, we should do one of three things if a group isn't specified:

1. Give an error that says "su requires both a user and a group, but only a user was specified", or something to that effect
2. Default to using the specified user's primary group
3. Just leave the egid alone and only switch the euid

If you let me know which approach you prefer, I can make a PR for it.

josephcsible:
Currently, su will happily accept a nonexistent user or group, but will then fail when it goes to set the euid/egid. Instead, catch this and fail early, to make the problem more obvious from the error message.

### DIFF
--- a/config.c
+++ b/config.c
@@ -1117,6 +1117,16 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
                                     "su\n", configFile, lineNum);
                             RAISE_ERROR();
                         }
+                        else if (newlog->suUid == NO_UID) {
+                            message(MESS_ERROR, "%s:%d no user for "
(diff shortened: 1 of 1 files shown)


josephcsible:
The build failure appears to be unrelated to my change.

kdudka:

PR #278, opened 2019-10-11
draft: re-inforce logrotate

cgzones:
- Silence/fix several issues/warnings from clang/scan-build/cppcheck/coverity
- Consistently check for memory allocation failures
- Tighten state and config file handling

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -110,9 +110,9 @@ install:
 script:
   - $COMPILER --version
   - ./autogen.sh
-  - ./configure --enable-werror CC=$COMPILER
+  - ./configure --enable-werror CC=$COMPILER --disable-silent-rules
   - make
-  - make -j9 check
+  - make -j9 check || ( cat test/test-suite.log; exit 1; )
   - make -j9 distcheck
   # nodeps because rpm build deps can not be installed on debian system
   - make rpm RPM_FLAGS="--nodeps"
--- a/Makefile.am
+++ b/Makefile.am
@@ -10,7 +10,7 @@
 #
 AM_CPPFLAGS = -include config.h
 sbin_PROGRAMS = logrotate
-logrotate_SOURCES = config.c log.c logrotate.c \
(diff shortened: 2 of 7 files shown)


.travis.yml
```diff
@@ -111,12 +111,12 @@ script:
   - $COMPILER --version
   - ./autogen.sh
   - ./configure --enable-werror CC=$COMPILER
-  - make
-  - make -j9 check
-  - make -j9 distcheck
+  - make V=1
+  - make V=1 -j9 check || cat test/test-suite.log
```
glensc: the `|| cat` breaks `set -e` enforcement and break travis build at that command. you may want to have:

```
  - make V=1 -j9 check || { cat test/test-suite.log; exit 1; }
```

.travis.yml
```diff
@@ -111,12 +111,12 @@ script:
   - $COMPILER --version
   - ./autogen.sh
   - ./configure --enable-werror CC=$COMPILER
-  - make
-  - make -j9 check
-  - make -j9 distcheck
+  - make V=1
```
glensc: instead of repeating `V=1` everywhere, would `export MAKEFLAGS=V=1` achieve the same goal?

- https://www.gnu.org/software/autoconf/manual/autoconf-2.61/html_node/The-Make-Macro-MAKEFLAGS.html

cgzones:
Yes, I will resubmit this pr with the changes incorporated.
Thanks for the feedback.

kdudka:
@cgzones I would prefer to finalize the review process for the originally submitted changes first and then review all the other changes in separate PRs.  This PR already tends to be messy...

Anyway, to be clear, I am not convinced that the extremely complex code of questionable quality and license constraints in `common.c` ever belongs to logrotate.

kdudka:
This seems to get stalled.  @cgzones Are you going to resubmit the pull request or should I move it forward myself?

cgzones:
Reworked in #280

PR #280, opened 2019-11-05
re-inforce logrotate (nr2)

cgzones:
* Silence/fix several issues/warnings from clang/scan-build/cppcheck/coverity
* Consistently check for memory allocation failures

Follow-up of #278 
Dropped the rewrite of state and config file handling


kdudka:
Thanks!  Merging everything, except the following commits:

  * c2c56654f3840752e04e1a5552a889c6dd253f1c - The `WUR` macros all over the code looked too disruptive and non-standard to me.
  * 466967245e07b905e6979be3426608bfc4a4b44c - It was a step backwards.  We are improving logrotate to be clear in its diagnostic messages and exit codes.  If a `MESS_ERROR` diagnostic message is printed, it should exit with non-zero exit code.  See related PR #239.

cgzones:
Regarding

> 4669672 - It was a step backwards. We are improving logrotate to be clear in its diagnostic messages and exit codes. If a MESS_ERROR diagnostic message is printed, it should exit with non-zero exit code. See related PR #239.

The current code is
```
[...]
unlink_file:
    if (unlink(filename) == 0)
        return 0;
    if (errno != ENOENT)
        return 1;

    /* unlink of log file that no longer exists is not a fatal error */
    message(MESS_ERROR, "error unlinking log file %s: %s\n", filename,
            strerror(errno));
    return 0;
}
```

* If `unlink()` succeeds, nothing is printed and zero is returned.
  Ok.
* If `unlink()` fails with `ENOENT`, an error is printed and zero is returned.
  This does not match.
* If `unlink()` fails except with `ENOENT`, nothing is printed and one is returned.
   This also does not match.

kdudka:
You are right that it is already messy.  I think that the diagnostic message in question should be turned into `MESS_WARNING` once we introduce it.  The real error message is now printed by `removeLogFile()` if `shred_file()` fails, but nothing is printed by `compressLogFile()` if `shred_file()` fails.

PR #282, opened 2019-11-19
state file handling

cgzones:
* move creation of stub state file to writeState
  Currently readState() creates a stub state file if none exists already.
  This stub file is only used later in writeState() to compute the file attributes for the final state file.

  Move the creation from readState() to writeState() to merge this logic.

  Also use the exclusive flag to create the stub file, to avoid a theoretical toctou race condition.
* readState: open state file only once
* writeState: fail if old temporary state file can not be removed
* writeState: improve error message
* writeState: delay unlink to not alter errno

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -2632,8 +2632,8 @@ static int writeState(const char *stateFilename)
         if (rename(tmpFilename, stateFilename)) {
             unlink(tmpFilename);
             error = 1;
-            message(MESS_ERROR, "error renaming temp state file %s\n",
-                    tmpFilename);
+            message(MESS_ERROR, "error renaming temp state file %s to %s\n",
+                    tmpFilename, stateFilename);
         }
     }
     else {
--- a/logrotate.c
+++ b/logrotate.c
@@ -2451,7 +2451,14 @@ static int writeState(const char *stateFilename)
     strcpy(tmpFilename, stateFilename);
     strcat(tmpFilename, ".tmp");
     /* Remove possible tmp state file from previous run */
-    unlink(tmpFilename);
+    error = unlink(tmpFilename);
+    if (error == -1 && errno != ENOENT) {
+        message(MESS_ERROR, "error removing old temporary state file %s: %s\n",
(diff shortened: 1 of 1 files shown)


logrotate.c
```diff
@@ -2640,22 +2673,6 @@ static int readState(const char *stateFilename)
         }
     }
 
-    if (!debug && (f_stat.st_size == 0)) {
-        /* create the file before continuing to ensure we have write
-           access to the file */
```
kdudka: I am afraid that the comment you removed had been there for a reason.  Should we trigger any rotation if we know that we will not be able to write that state file afterwards?

cgzones: This might be a valid concern, but currently we rotate regardless of having write access or not
https://github.com/logrotate/logrotate/blob/44a097a227eb28252424fe0f2281b4e97d06169d/logrotate.c#L2940

kdudka: Indeed.  Sorry for missing it!

logrotate.c
```diff
@@ -2451,15 +2451,51 @@ static int writeState(const char *stateFilename)
...
+                      S_IWUSR | S_IRUSR | S_IRGRP);
+        if (fdcurr == -1) {
+            message(MESS_ERROR, "error creating stub state file %s: %s\n",
+                    stateFilename, strerror(errno));
+            free(tmpFilename);
+            return 1;
+        }
+
+        /* check write access */
+        ret = write(fdcurr, state_header_v2, state_header_v2_len);
```
kdudka: Please use `full_write()` instead of `write()`.  `fopen/fprintf/fclose` is not really equivalent to `open/write/close`.

kdudka:
As for the state file permissions, I propose to use `0644`.  I think there is no secret information in the state file to deny read access to it by default.

logrotate.c
cgzones: changed, thanks

cgzones: changed, thanks

kdudka:

Issue #288, opened 2020-01-08
compilation will fail with gcc 10


scarabeusiv:
Gcc 10 introduces new behavior with -fno-common and the package fails to build.

Explanation of the change can be found at our downstream bug https://bugzilla.suse.com/show_bug.cgi?id=1160244

I didn't check the master, but release fails as bellow:
[logrotate.txt](https://github.com/logrotate/logrotate/files/4034201/logrotate.txt)

PR #289, opened 2020-01-08
split declaration and definition of queue variable

cgzones:
Support compilation with -fno-common flag, which is the default for GCC 10.

Fixes: #288


kdudka:
Merging, thanks!

PR #290, opened 2020-01-08
travis: add scan-build job

cgzones:
Also bump for analysis jobs the distro to bionic and clang to version 9.0


kdudka:
Merging, thanks!

PR #291, opened 2020-01-08
logrotate.service: add systemd hardening options

cgzones:
+ RestrictNamespaces : prohibit namespacing
+ LockPersonality : prohibit usage of personality(2)
+ ProtectKernelLogs : prohibit access to the kernel ring buffer (syslog(2) (NOT syslog(3)), /dev/kmsg and /proc/kmsg)


kdudka:
Merging, thanks!

PR #294, opened 2020-02-05
config.c: fixing the alloca.h include for OpenBSD

rom1sqr:
Currently working on packaging logrotate for OpenBSD (via ports).
As any other *BSD system, the alloca() function is defined in
stdlib.h, so no need to include alloca.h (does not exist anyway).


kdudka:
Merging, thanks!

Issue #295, opened 2020-02-05
Support locking / prevent concurrent compress jobs of the same log file


raoulbhatia:
Hi,

I am encountering duplicate runs of gzip related to big log files that need rotating when calling `logrotate` i.e. hourly.

My use case:
* I am receiving 100-500GB of logs via `rsyslog` (daily) that I save to a local logfile.
* To ensure that logs are rotated in i.e. sizeable junks, I run `logrotate` hourly.
* To save on disk space, i enabled `delaycompress`.
* When the system is under load (CPU and/or IO), compression might take >1hr.  Thus, logrotate is started again and ends up `delaycompress`-ing the same file again, contributing to the overload of the system.  In this way, I sometimes see 4-5 concurrent instances of `logrotate` & `gzip`.

Is there a way to handle this (w/o implementing a locking mechanism myself), or would this require new functionality in logrotate?

Thanks,
Raoul

kdudka:
If it happens that more processes of logrotate run in parallel on the same status file and log files, there will be much more issues than running the compressor in parallel.  In general, logrotate does not support parallel invocation unless it operates on separate status file and disjoint set of log files.

So it would be necessary to lock the whole run of logrotate.  What should be the expected action when logrotate is triggered while the previous process is still running?  Should it be skipped or queued indefinitely?

raoulbhatia:
Thanks for getting back to me quickly.  I can think of the following locking solutions:

* Per **logfile** that is currently processed.
  * There is already some logic for compress that handles situations where `logfile.1` exists?  Perhaps this can be extended?
  * Perhaps introduce a .lck suffix for the file that is processed?
* Per **logrotate section**, i.e. `/var/log/news/* { [...] }`
* Per **logrotate config file**, i.e. `/usr/sbin/logrotate /opt/logrotate1.conf` will not lock `/usr/sbin/logrotate logrotate2.conf`.
  * However, I could think of potential issues for `/usr/sbin/logrotate /etc/logrotate.conf` vs. `/usr/sbin/logrotate /etc/logrotate.d/myconf`.
* Per **state file** (least preferred, because one specific config/logfile could lock everything else)

I also see a potential challenge when logrotate might be killed, either manually or perhaps by the system itself.  Then the locking will prevent further runs.  Perhaps there is a solution to specify "locking is only valid for 1 day", or "check if another logrotate pid has this file open"?

kdudka:
[...]

PR #297, opened 2020-02-06
lock state file to prevent parallel execution

cgzones:
related to #295

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -12,6 +12,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/file.h>
 #include <sys/stat.h>
 #include <sys/wait.h>
 #include <time.h>
@@ -2511,37 +2512,12 @@ static int writeState(const char *stateFilename)
 
     fdcurr = open(stateFilename, O_RDONLY);
     if (fdcurr == -1) {
-        const char *state_header_v2 = "logrotate state -- version 2\n";
-        const size_t state_header_v2_len = strlen(state_header_v2);
-        size_t ret;
-
-        /* no error if state file is just missing */
-        if (errno != ENOENT) {
-            message(MESS_ERROR, "error opening state file %s: %s\n",
-                    stateFilename, strerror(errno));
-            free(tmpFilename);
-            return 1;
-        }
-
-        /* create a stub state file with mode 0644 */
-        fdcurr = open(stateFilename, O_CREAT | O_EXCL | O_WRONLY,
-                      S_IWUSR | S_IRUSR | S_IRGRP | S_IROTH);
-        if (fdcurr == -1) {
-            message(MESS_ERROR, "error creating stub state file %s: %s\n",
-                    stateFilename, strerror(errno));
-            free(tmpFilename);
-            return 1;
-        }
-
-        /* check write access */
-        ret = full_write(fdcurr, state_header_v2, state_header_v2_len);
-        if (ret != state_header_v2_len) {
-            message(MESS_ERROR, "error writing to stub state file %s: %s\n",
-                    stateFilename, strerror(errno));
-            close(fdcurr);
-            free(tmpFilename);
-            return 1;
-        }
+        /* the statefile should exist, lockState() already created an empty
+         * state file in case it did not exist initially */
+        message(MESS_ERROR, "error opening state file %s: %s\n",
+                stateFilename, strerror(errno));
+        free(tmpFilename);
+        return 1;
     }
 
     /* get attributes, to assign them to the new state file */
@@ -2721,12 +2697,15 @@ static int readState(const char *stateFilename)
         /* treat non-openable file as an empty file for allocateHash() */
         f_stat.st_size = 0;
 
-        /* no error if state just not exists */
-        if (errno != ENOENT) {
-            message(MESS_ERROR, "error opening state file %s: %s\n",
-                    stateFilename, strerror(errno));
+        /* the statefile should exist, lockState() already created an empty
+         * state file in case it did not exist initially */
+        message(MESS_ERROR, "error opening state file %s: %s\n",
+                stateFilename, strerror(errno));
 
-            /* do not return until the hash table is allocated */
+        /* Do not return until the hash table is allocated.
+         * In debug mode the state file might not exist,
+         * cause lockState() is not called */
+        if (!debug) {
             rc = 1;
         }
     } else {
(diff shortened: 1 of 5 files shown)


kdudka:
Thank you for working on this!  I will review it after the planned release of logrotate.  My initial feeling is that this will need to be optional to cover the environments where the locking mechanism does not work reliably.  There is some risk that the change will trigger problems even without parallel invocation.

logrotate.c
```diff
@@ -2484,6 +2485,14 @@ static int writeState(const char *stateFilename)
             return 1;
         }
 
+        if (flock(fdcurr, LOCK_EX | LOCK_NB) == -1) {
```
kdudka: Why do we care about locking in `writeState()`?  Isn't it too late?

Some log files might have been already rotated at this point.  Why should we prevent the status file from being written in case we are not able to acquire the lock?

logrotate.c
```diff
@@ -2860,6 +2869,36 @@ static int readState(const char *stateFilename)
     return 0;
 }
 
+static int lockState(const char *stateFilename)
+{
+    int lockFd = open(stateFilename, O_RDWR);
```
kdudka: Shouldn't we open the file with `O_CLOEXEC` to avoid leaking the file descriptor to subprocesses?

logrotate.c
```diff
@@ -2860,6 +2869,36 @@ static int readState(const char *stateFilename)
...
+    if (lockFd == -1) {
+        if (errno == ENOENT) {
+            return 0;
```
kdudka: Does it mean we will proceed with rotations without holding any lock in this case?

logrotate.c
```diff
@@ -2978,6 +3017,10 @@ int main(int argc, const char **argv)
     poptFreeContext(optCon);
     nowSecs = time(NULL);
 
+    if (lockState(stateFile)) {
+        exit(3);
```
kdudka: Why exactly 3 was chosen as the exit code here?  This should be documented in any case.

test/test-0085.sh
```diff
@@ -0,0 +1,22 @@
...
+# check state file locking
+cleanup 85
+
+preptest test.log 85 1
+
+touch state
+
+$RLR test-config.85 -v -f &
+
+sleep 1
```
kdudka: Waiting 1 second for logrotate to start in the background might not be sufficient in case the tests are runing in parallel on an overloaded build host.

logrotate.c
cgzones: This is just paranoid locking of the temporary state file, shouldn't matter much though.

logrotate.c
cgzones: yep, added

logrotate.c
cgzones: Now `lockState()` creates an empty stub statefile in that case

logrotate.c
cgzones: The exit values 1 and 2 are used, so I wanted to make it distinct.
But this is subject to changes and yes we should document it.

test/test-0085.sh
cgzones: Increased the timeout values (waiting for the first run to acquire the lock and how long the first run will hold the lock)

logrotate.c
kdudka: In which scenario exactly would the code help?  I still think that the state file should be written.  When the logs have already been rotated, not reflecting it in the state file could hardly improve anything.

test/test-0085.sh
kdudka: Fair enough.  We can implement explicit synchronization later if needed.

logrotate.c
```diff
@@ -2720,14 +2729,12 @@ static int readState(const char *stateFilename)
         /* treat non-openable file as an empty file for allocateHash() */
         f_stat.st_size = 0;
 
-        /* no error if state just not exists */
-        if (errno != ENOENT) {
-            message(MESS_ERROR, "error opening state file %s: %s\n",
-                    stateFilename, strerror(errno));
+        /* the statefile shoud exist, lockState() is holding a lock */
```
kdudka, line 2732: This is misleading.  Holding the lock does not anyhow prevent `stateFilename` from being unlinked.  It should rather say "lockState() already created an empty state file in case it did not exist initially".

logrotate.c
cgzones: Changed to fail if state file does not exist at this point

logrotate.c
cgzones: Improved comment

logrotate.c
kdudka: Looks good. Thanks for the update!

logrotate.c
```diff
@@ -2909,6 +2886,51 @@ static int readState(const char *stateFilename)
     return 0;
 }
 
+static int lockState(const char *stateFilename)
+{
+    int lockFd = open(stateFilename, O_RDWR | O_CLOEXEC);
```
kdudka, line 2891: With `debug == 1` we either need to open the state file with `O_RDONLY`, or rather skip the locking entirely.  Otherwise we break the use case where a non-privileged runs `logrotate -d /etc/logrotate.conf`:

With current master (7da0c415da2ad1f56b67cbdf7ad159a9fe5e6d2a)
```
$ ./logrotate -d /etc/logrotate.conf
WARNING: logrotate in debug mode does nothing except printing debug messages!  Consider using verbose mode (-v) instead if this is not what you want.

reading config file /etc/logrotate.conf
including /etc/logrotate.d
Ignoring yum.rpmsave, because of *.rpmsave pattern match
reading config file bootlog
reading config file btmp
reading config file chrony
reading config file cyrus-imapd
reading config file dnf
reading config file dpkg
reading config file fedmsg
reading config file glusterfs
reading config file httpd
reading config file iscsiuiolog
reading config file libvirtd
reading config file libvirtd.qemu
reading config file numad
reading config file ppp
reading config file squid
reading config file tinyproxy
reading config file vsftpd
reading config file wtmp
reading config file yum
Reading state from file: /var/lib/logrotate.status
Allocating hash table for state file, size 64 entries
[...]
```

With this pull request:
```
$ ./logrotate -d /etc/logrotate.conf
WARNING: logrotate in debug mode does nothing except printing debug messages!  Consider using verbose mode (-v) instead if this is not what you want.

reading config file /etc/logrotate.conf
including /etc/logrotate.d
Ignoring yum.rpmsave, because of *.rpmsave pattern match
reading config file bootlog
reading config file btmp
reading config file chrony
reading config file cyrus-imapd
reading config file dnf
reading config file dpkg
reading config file fedmsg
reading config file glusterfs
reading config file httpd
reading config file iscsiuiolog
reading config file libvirtd
reading config file libvirtd.qemu
reading config file numad
reading config file ppp
reading config file squid
reading config file tinyproxy
reading config file vsftpd
reading config file wtmp
reading config file yum
error: error opening state file /var/lib/logrotate.status: Permission denied
```

cgzones: True, only tested in debug mode with an non-existent state file...

Do we need an opt-out option (`--skip-state-lock`)?

kdudka: The fix looks good, thanks!

As for `--skip-state-lock`, what should be the use case?

If one needs to enforce rotation while there is a stray process of logrotate, they should be able to either kill the process or replace the state file.

If we want to cover platforms where `flock()` is not available, we should probably consider a compile-time check instead.

The only case that comes to my mind is a file system where `flock()` would fail at run-time.

Is there any other use case that I am overseeing?

Anyway, the option is easy to implement and does not constitute any maintenance or administration burden.  I do not have strong opinion about it...

cgzones: Added the option `--skip-state-lock`.

If `flock(2)` is unavailable, compilation will fail, so packagers/user will stumble over it upfront and can tackle it.

I was considering systems where for whatever reason locking in unsupported at runtime or prohibited. At that time logrotate is compiled/shipped and there would be no easy way (other then re-compiling, what might be difficult or prohibited at that system) to get around.

kdudka: Looks good.  Thanks for the update!

kdudka:
Merging now.  Thank you for being patient while contributing this!

PR #302, opened 2020-02-26
manpage: move maxage configuration option to Rotation section

cgzones:


kdudka:
Sorry for the delay!  I am not convinced this change goes the right direction.  `maxage` is about file selection, isn't it?  It selects files to be removed.  Moreover, I like that `minage` and `maxage` are currently next to each other.

cgzones:
fair enough

PR #305, opened 2020-03-13
silence unused parameter warning in acl disabled build

cgzones:


kdudka:
This is apparently a typo from commit a56a4f04b2351725ede76dead56af36c2fb2ac14.  The actual fixup should rather look like this:

```diff
--- a/logrotate.c
+++ b/logrotate.c
@@ -588,7 +588,7 @@ static int createOutputFile(const char *fileName, int flags, const struct stat *
         }
     }
 #else
-    (void) acl_set;
+    (void) acl;
 #endif

     if (!acl_set || force_mode) {
```

Do you agree?

cgzones:
Good catch!
`acl_set` is used unconditionally right after the `#endif`

kdudka:
Merging, thanks!

Issue #301, opened 2020-02-26
Maxage doesn't delete old logs without using dateext


swoopdk:
I seem to have run into an issue. 
When using something like this example the `maxage` option doesn't delete the older logfiles (regardless of how old the mtime on them are) unless we also use the `dateext` option. 

Shouldn't it be working regardless if `dateext` is being used or not ? 

I should note i'm using the `rotate -1` option to only have older files be deleted on age and irregardless of number of logs.

Example of rotate.conf that does not seem to remove older logs ever. 

```
/home/swoopdk/logtest/logs/*.log {
  daily
  rotate -1
  ifempty
  maxage 2
}
```

However this one seems to work from my initial testing. 

```
/home/swoopdk/logtest/logs/*.log {
  dateext
  daily
  rotate -1
  ifempty
  maxage 2
}
```

I have not looked into the code yet and might be fixed in upcomming releaese this friday, but from what i gather its a bug. Unless i'm misreading the documentation and the `rotate -1` doesn't do what i think it does.

cgzones:
I can confirm that there is a diverging behavior.

The question is: what should happen when using `rotate -1` and `maxage`?
  * no removal
       man-page says: *If count is -1, old logs are not removed at all*
  * removal

Things to do:
  * unify behavior whether using `dateext` or not
  * document in the man-page which configuration takes precedence: `rotate -1` or `maxage`
  * add test-case

swoopdk:
From what me and colleagues read in documentation we were expecting the following behavior

`rotate -1` should say no removal at all **unless** something else like maxage wants to remove log files

This would allow logrotate the very dynamic settings og allowing logs to be removed solely based on age and not number of logs. 

Regarding the `rotate 100000`case it's not applicaple for my specific use case atm for example. I need to keep all logs for X years BUT i also need to cover that case that logs might be rotated many times a day. So the rotate count would skyrocket. 

From my initial impression of https://github.com/logrotate/logrotate/issues/202 i would support our way of thinking that `rotate -1` disabled deletion but still would allow example `maxage` to remove. 

Of course our workaround is to run a find command in post-rotate but i think as a tool logrotate would benefit from having each option work independently. After all when they are both enabled they work together that either maxage **OR** rotate is the trigger for deletion right ?

But super happy to hear it's not just me going nuts here and hope we can support my own usecase as it would make managing our logs greatly more elegant than using pre and post rotate scripts to do most of the labor :)

cgzones:
[...]

PR #307, opened 2020-03-16
Delete old logs hit by maxage regardless of dateext (Fix: #301)

cgzones:
Currently old logs hit by `maxage` are only removed if `dateext` is used.
Unify the behavior to remove them.

Document the relation between `maxage` and `rotate -1` in the manpage.

Fixes: #301


kdudka:
Merging, thanks!

PR #308, opened 2020-03-19
Print compressoptions

cgzones:


logrotate.c
```diff
@@ -752,7 +752,7 @@ static int compressLogFile(const char *name, const struct logInfo *log, const st
     char *prevCtx;
     pid_t pid;
 
-    message(MESS_DEBUG, "compressing log with: %s\n", log->compress_prog);
+    message(MESS_DEBUG, "compressing log with %s: %s\n", compress_options_count ? "added options" : "default", log->compress_prog);
```
kdudka: This should have been `log->compress_options_count` I guess.

logrotate.c
```diff
@@ -767,7 +767,7 @@ static int compressLogFile(const char *name, const struct logInfo *log, const st
     sprintf(compressedName, "%s%s", name, log->compress_ext);
 
     if ((inFile = open(name, O_RDWR | O_NOFOLLOW)) < 0) {
-        message(MESS_ERROR, "unable to open %s for compression\n", name);
+        message(MESS_ERROR, "unable to open %s for compression: %s\n", name, strerror(errno));
```
kdudka: This change makes sense.

logrotate.c
```diff
@@ -752,7 +752,7 @@ static int compressLogFile(const char *name, const struct logInfo *log, const st
     char *prevCtx;
     pid_t pid;
 
-    message(MESS_DEBUG, "compressing log with: %s\n", log->compress_prog);
+    message(MESS_DEBUG, "compressing log with %s: %s\n", log->compress_options_count ? "added options" : "default", log->compress_prog);
```
kdudka: The updated message sounds confusing to me:

> compressing log with default: /bin/gzip

I would rather understand it as if /bin/gzip was the default compress program.

What is the motivation behind this change?

If we want to increase the verbosity, we should rather print the array passed to `execvp()`.  But is it really needed? 

The original string that the array is build from is already printed while parsing the configuration...

cgzones: Yeah, you're right; this is not needed.

kdudka:
Thanks for the quick update!  Merging...

PR #309, opened 2020-03-20
Remove redundant include of <limits.h>

cgzones:


kdudka:
How did you verify that the include is redundant on Solaris?

Which version of the system did you test it with?

There are systems where the include is needed for `PATH_MAX` definition:
https://patchwork.kernel.org/patch/7272951/

cgzones:
I didn't test it on Solaris.

`<limits.h>` is unconditionally included at
https://github.com/logrotate/logrotate/blob/796084bd81e1c48afda59b654c6bd5cddc266c7a/config.c#L1-L2
https://github.com/logrotate/logrotate/blob/796084bd81e1c48afda59b654c6bd5cddc266c7a/logrotate.c#L1-L6

kdudka:
Indeed.  I did not spot that the include was duplicated originally.  Merging, thanks!

PR #310, opened 2020-03-20
readConfigFile: fix usage of RAISE_ERROR in nested loops

cgzones:
RAISE_ERROR might unfold to a continue, which is bad if we are in a nested for/while loop.
Jump to a goto-label instead.


kdudka:
Merging, thanks!

PR #313, opened 2020-04-01
Clang 10

cgzones:
* Bump clang for analyser and sanitizer run to version 10
* Resolve clang-analyzer warning:
```
config.c:970:11: warning: Although the value stored to 'pw' is used in the enclosing expression, the value is never actually read from 'pw'
    if (!(pw = getpwuid(getuid()))) {
          ^    ~~~~~~~~~~~~~~~~~~
```

There is on thing bothering me: the condition in config.c:readConfigFile()
https://github.com/logrotate/logrotate/blob/184d07687f3b6870ef6d6da454536c6bfa12f076/config.c#L999-L1007

0. `pw` can't be `NULL`, that's easy (see 4th commit)
1. so the conditions simplifies to:
`sb.st_uid != ROOT_UID && (sb.st_uid != pw->pw_uid || pw->pw_uid != ROOT_UID)`

If the first `&&` argument is `true` (e.g. `sb.st_uid` differs from `ROOT_UID`), the second argument can not be `false`: `pw->pw_uid` can not be equal to `sb.st_uid` and simultaneously equal to `ROOT_UID` (by the first `&&` argument)

So basically the current check collapses to `sb.st_uid != ROOT_UID`, should it be `sb.st_uid != ROOT_UID && sb.st_uid != pw->pw_uid`??

I think commit f9c16896d15916eeccb2ce950dd1313deb4f512c was not quite accurate.

### DIFF
--- a/config.c
+++ b/config.c
@@ -998,7 +998,7 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
             return 0;
         }
 
-        if (sb.st_uid != ROOT_UID && (pw == NULL ||
+        if (sb.st_uid != ROOT_UID && (
                     sb.st_uid != pw->pw_uid ||
                     pw->pw_uid != ROOT_UID)) {
             message(MESS_ERROR,
--- a/config.c
+++ b/config.c
@@ -925,7 +925,6 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
     char **scriptDest = NULL;
     struct logInfo *newlog = defConfig;
     char *start, *chptr;
-    struct passwd *pw;
     struct stat sb;
     int state = STATE_DEFAULT;
     int logerror = 0;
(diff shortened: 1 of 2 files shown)


config.c
```diff
@@ -967,13 +966,16 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
         return 0;
     }
 
-    if (!(pw = getpwuid(getuid()))) {
-        message(MESS_ERROR, "Logrotate UID is not in passwd file.\n");
+    if (!getpwuid(getuid())) {
+        message(MESS_ERROR, "Can not find logrotate UID (%d) in passwd file: %s\n",
```
kdudka, line 970: Although I am not a native speaker, I believe that "can not" should rather be spelled as "cannot": https://www.grammarly.com/blog/cannot-or-can-not/

I see we have the same pattern in all the "can not fork" messages, which I did not spot earlier.

kdudka:
Merging, thanks!

Issue #317, opened 2020-04-02
Suspicious condition


cgzones:
Follow-up of #313 

https://github.com/logrotate/logrotate/blob/77e77576fa03d293d886c6537ade44ad31ab0bf7/config.c#L1001-L1009

```
sb.st_uid != ROOT_UID && (sb.st_uid != pw->pw_uid || pw->pw_uid != ROOT_UID)
```

If the first `&&` argument evaluates to `true` (i.e. `sb.st_uid` differs from `ROOT_UID`), the second argument can not be `false`: `pw->pw_uid` can not be equal to `sb.st_uid` and simultaneously be equal to `ROOT_UID` (by the first `&&` argument)

So basically the current check collapses to `sb.st_uid != ROOT_UID`, should it be `sb.st_uid != ROOT_UID && sb.st_uid != pw->pw_uid`?

I think commit f9c1689 was not quite accurate.

kdudka:
Assuming that `getpwuid(ROOT_UID)->pw_uid == ROOT_UID` is always true, the whole condition can be simplified as `(sb.st_uid != ROOT_UID)` and the call of `getpwuid(ROOT_UID)` can be optimized out completely.

PR #318, opened 2020-04-02
Simplify check of configuration file owner

cgzones:
If the first && argument evaluates to true (i.e. sb.st_uid differs from
ROOT_UID), the second argument cannot be false: pw->pw_uid cannot be
equal to sb.st_uid and simultaneously be equal to ROOT_UID (by the
first && argument).

Assuming that (getpwuid(ROOT_UID)->pw_uid == ROOT_UID) is always true,
simplify the condition to (sb.st_uid != ROOT_UID) and drop the call
of getpwuid(ROOT_UID).

Closes: #317

### DIFF
--- a/config.c
+++ b/config.c
@@ -974,8 +974,6 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
     }
 
     if (getuid() == ROOT_UID) {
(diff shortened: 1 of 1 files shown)


kdudka:
Merging, thanks!

PR #319, opened 2020-04-08
switch_user_permanently: skip switchback check if switched to root

cgzones:
*  switch_user_permanently: skip switchback check if switched to root
   Allow switching only the real group (not the user) with a configuration
   like `su root somegroup`.
   E.g. mailman uses `su root list`, which currently fails with:
       `error: failed to switch user permanently, able to switch back`
      `error: failed to compress log /var/log/mailman/qrunner.1`

*  switch_user_permanently: add sanity check that effective ids match configuration specified ones

*  switch_user*: improve debug logging
   Print pid to distinguish processes.
   Print previous effective ids.

* rotateLogSet: call switch_user_back on early return

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -2317,6 +2317,9 @@ static int rotateLogSet(const struct logInfo *log, int force)
 
     if (state == NULL || rotNames == NULL) {
         message_OOM();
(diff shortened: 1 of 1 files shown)


kdudka:
Merging, thanks!

Issue #321, opened 2020-05-05
Order of global and include directives in config


cgzones:
Stanislav Sukholet reported at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=959773 that global options, like compress, are applied only to subsequent `include` directives.

E.g.
```
include /etc/logrotate.d
compress
```
vs 
```
compress
include /etc/logrotate.d
```

Changing the current behaviour might break usages like:
```
include /etc/logrotate.standard.d
mail root@localhost
include /etc/logrotate.mail.d
```

Should we add a note to the configuration like:
```
---
 examples/logrotate.conf | 3 +++
 1 file changed, 3 insertions(+)

diff --git a/examples/logrotate.conf b/examples/logrotate.conf
index d5ab23f..987d4ab 100644
--- a/examples/logrotate.conf
+++ b/examples/logrotate.conf
@@ -1,4 +1,7 @@
 # see "man logrotate" for details
+
+# global options affect only subsequent include directives
+
 # rotate log files weekly
 weekly
```
?

kdudka:
Sounds like a good idea to me.  Should we also mention it in the man page?

cgzones:
Probably can't hurt

PR #325, opened 2020-05-15
logrotate.conf: mention to order global options first

cgzones:
Options defined after an include directive won't apply to that statement.

Closes: #321


kdudka:
Sorry for the delay.  On a second read I find the statement confusing.  One could understand it as "global options affect nothing else but included configuration files", which is obviously not the case.

Would not it be safer to word it the other way around?

"global options do not affect preceding include directives"

kdudka:
Thanks for the update!  Merging...

Issue #322, opened 2020-05-15
include directive can't use tilde


dvzrv:
Hi! I've tried using a systemd user unit for the setup of a per-user logrotate.

This seems to work fine, but the `include` directive seems to not accept tilde to resolve a directory.
When defaulting to some sort of [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), we can assume, that e.g. `~/.config/logrotate.conf` holds the per user configuration, and that we would be able to place per application configuration files to `~/.config/logrotate.d/`.
However, this does not get resolved by the include:

```
including ~/.config/logrotate.d
error: cannot stat ~/.config/logrotate.d: No such file or directory
```
(but the directory exists and has a configuration snippet in it)

Using the rotate definitions directly in the configuration file works though (here with a sway log as an example):

```
rotating pattern: ~/.log/sway.log  weekly (4 rotations)
empty log files are not rotated, old logs are removed
considering log /home/dave/.log/sway.log
Creating new state
Now: 2020-05-15 09:43
Last rotated at 2020-05-15 09:00
log does not need rotating (log has already been rotated)
```

It would be really great if the include directive could also resolve tilde, as the only alternative is placing everything into one configuration file, or hard coding the `include` directory (which then is not portable to another system where e.g. the user name might be different).

cgzones:
Something like this: #326 ?

PR #326, opened 2020-05-15
Allow to pass a home-relative path to include

cgzones:
Accept a path with leading "~/" in include directives and replace with the environment variable ${HOME} (if existent).

### DIFF
--- a/config.c
+++ b/config.c
@@ -1478,6 +1478,39 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
                                            &buf, length);
                         if (key == NULL)
                             continue;
+
+                        if (key[0] == '~' && key[1] == '/') {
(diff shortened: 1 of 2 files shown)


kdudka:
This definitely needs to be documented.  Should we take the value from `getpwuid(getuid())->pw_dir` in case `$HOME` is unset to better emulate the tilde expansion in shell?

dvzrv:
@cgzones thanks for looking into this and following up on #322.

I'm not familiar with the code base, but this looks like it will do what I envisioned in the issue.

cgzones:
Updated.

Can you please either mention to keep the code ISO C90 compliant, e.g. in `CONTRIBUTING.md`, or add the warning `-Wdeclaration-after-statement`?

kdudka:
@cgzones Pull request #328 enables the warning.

PR #334, opened 2020-06-23
Add check that copytruncate and rotate 0 does not keep a rotated log file

cgzones:
Fixed in version 3.15


kdudka:
Merging, thanks!

PR #344, opened 2020-07-10
use getenv on macOS and BSD

SeekingMeaning:
`secure_getenv` is not available on non-GNU systems (such as macOS)

### DIFF
--- a/config.c
+++ b/config.c
@@ -45,6 +45,10 @@ struct logInfoHead logs;
 #include "asprintf.c"
 #endif
 
+#if !defined(HAVE_SECURE_GETENV)
+#define secure_getenv getenv
+#endif
+
 #if !defined(HAVE_ASPRINTF) && !defined(_FORTIFY_SOURCE)
 #include <stdarg.h>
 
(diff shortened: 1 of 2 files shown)


cgzones:
We should probably add `secure_getenv` to `AC_CHECK_FUNCS` in `configure.ac` and then toggle at compile time based on `HAVE_SECURE_GETENV`

SeekingMeaning:
_Challenge accepted_

SeekingMeaning:
Which would you prefer?

```c
#if defined(HAVE_SECURE_GETENV)
                            const char *env_home = secure_getenv("HOME");
#else
                            const char *env_home = getenv("HOME");
#endif
```

```c
#if !defined(HAVE_SECURE_GETENV)
#define secure_getenv getenv
#endif
```

```c
#if defined(HAVE_SECURE_GETENV)
#define getenv secure_getenv
#endif
```

cgzones:
My preference is: 

```
#if !defined(HAVE_SECURE_GETENV)
#define secure_getenv getenv
#endif
```

Also please keep `AC_CHECK_FUNCS` sorted

cgzones:
Thanks,
LGTM :+1: 

(Maybe the commits can be squashed before merging)

kdudka:
Merging, thanks!

PR #345, opened 2020-07-10
logrotate.service: harden with ProtectHostname and RestrictSUIDSGID

cgzones:
Prohibit changing the system hostname/domain and creating SUID/SGID files

### DIFF
--- a/examples/logrotate.service
+++ b/examples/logrotate.service
@@ -18,12 +18,14 @@ IOSchedulingPriority=7
 #  no ProtectHome for userdir logs
 #  no PrivateNetwork for mail deliviery
 #  no NoNewPrivileges for third party rotate scripts
+#  no RestrictSUIDSGID for creating setgid directories
 LockPersonality=true
 MemoryDenyWriteExecute=true
 PrivateDevices=true
 PrivateTmp=true
 ProtectClock=true
 ProtectControlGroups=true
+ProtectHostname=true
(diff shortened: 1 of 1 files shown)


examples/logrotate.service
```diff
@@ -24,9 +24,11 @@ PrivateDevices=true
...
 ProtectClock=true
 ProtectControlGroups=true
+ProtectHostname=true
 ProtectKernelLogs=true
 ProtectKernelModules=true
 ProtectKernelTunables=true
 ProtectSystem=full
 RestrictNamespaces=true
 RestrictRealtime=true
+RestrictSUIDSGID=true
```
kdudka, line 34: I am not sure about this one.  Setting setuid/setgid bits on directories is quite common.  What is the motivation behind this change?

cgzones: Is it really common to set setuid/setgid bits inside of logrotate scripts?

kdudka: Who are you asking here?  Users of logrotate?  This is not the channel to reach them...

Note that these bits are not limited to scripts only.  They can be set by logrotate directly, for example: `createolddir 6770`

cgzones: OK, I understand now.
Converted into a comment why not using RestrictSUIDSGID

kdudka:
Merging, thanks!

PR #346, opened 2020-07-10
test-70: fix minage test in reprotest environment

cgzones:
The minage condition is based on the mtime of the log file,
not on the rotation date from the state file.
During reproducibility test, which changes the system time, this test
may fail due to the file actually being to old.
This test 70 should assure that we do not rotate, actual rotation is
tested in test 71.


cgzones:
I am not sure any more that this is the right approach. I'll try to debug further...

PR #356, opened 2020-09-17
travis: add FreeBSD job and resolve portability issues

cgzones:


test/test-0034.sh
```diff
@@ -10,9 +10,12 @@ cleanup 34
 # checks the logrotate -d output.
 preptest test.log 34 1 0
 
-$RLR test-config.34 -d -f 2>&1|grep "uid = 0 gid = 0" > /dev/null
+OUTPUT=$($RLR test-config.34 -d -f 2>&1)
+
+echo ${OUTPUT} | grep "uid = 0 gid = 0" > /dev/null
```
kdudka, line 15: I would expect `${OUTPUT}` to be quoted, so that grep does not get one overly long line on its input.  But it probably does not affect the functionality in this case.

kdudka (approved):

test/test-0034.sh
cgzones: Added quotes

Issue #190, opened 2018-03-30
Feature request: copy_punch_hole rotation mode


tailhook:
Hi,

I've came across the wild idea of using `FALLOC_FL_PUNCH_HOLE` for log rotation.

The idea is that if we use `copytruncate` mode, there is a small window where messages can be lost. But if we copy log and instead of truncating the file do `fallocate(..FALLOC_FL_PUNCH_HOLE..)` of exactly the copied part we still free the disk space (even if the virtual file size is still large) and never loose any writes.

What do you think? Is it even viable? Should I make a PR?

kdudka:
How would it work for subsequent rotations?  Where would you keep track of the blocks that were deallocated?

tailhook:
Probably use `SEEK_HOLE` (or maybe `FS_IOC_FIEMAP`) ?

kdudka:
I am afraid that such a feature would be highly dependent on kernel version and implementation of a particular file system.  Nevertheless, if the feature is optional and works reliably on a reasonable sample of commonly used (file) systems, I am fine with that.

CAFxX:
I think this would be even possible to enable transparently in `copytruncate` mode if the feature is detected at runtime as supported by the filesystem. Assuming it works reliably it would eliminate the potential loss of logs that happen between copy and truncate.

PR #361, opened 2020-10-11
Add copyreduce mode

yosshy:
copytruncate can loss some logging data because it may take time
between copying file and truncating it. copyreduce is like
copytruncate but it uses fallocate() instead of ftruncate() to reduce
source file size without any loss of logging data.

It reduces a source file size in 2 steps below:

1. fallocate(FALLOC_FL_COLLAPSE_RANGE) from 0 to
   (filesize / blocksize) * blocksize) bytes to truncate most of
   the file blocks because we can only use FALLOC_FL_COLLAPSE_RANGE
   per block.

2. fallocate(FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE) from 0 to
   (filesize - ((filesize / blocksize) * blocksize)) bytes to make a
   punch hole for the rest part of 1. Therefore, reading it returns
   NULL characters.

Linux 3.15+ supports fallocate(FALLOC_FL_COLLAPSE_RANGE) on XFS and
ext4, so we can use copyreduce mode on them.

Fixes: https://github.com/logrotate/logrotate/issues/190


config.c
```diff
@@ -1101,6 +1101,12 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
                         newlog->flags |= LOG_FLAG_DATEYESTERDAY;
                     } else if (!strcmp(key, "datehourago")) {
                         newlog->flags |= LOG_FLAG_DATEHOURAGO;
+#if defined(WITH_FALLOCATE)
+                    } else if (!strcmp(key, "copyreduce")) {
+                        newlog->flags |= LOG_FLAG_COPYREDUCE;
+                    } else if (!strcmp(key, "nocopyreduce")) {
+                        newlog->flags &= ~LOG_FLAG_COPYREDUCE;
+#endif
```
cgzones, line 1109: maybe use a different error message if used on an unsupported platform than the generic message?
e.g.:
```c
#else
    } else if (!strcmp(key, "copyreduce") || !strcmp(key, "nocopyreduce")) {
        message(MESS_ERROR, "%s:%d unknown option '%s' "
                "-- ignoring line\n", configFile, lineNum, key);
        if (*start != '\n')
            state = STATE_SKIP_LINE;
#endif
```

logrotate.c
```diff
@@ -1221,6 +1226,28 @@ static int copyTruncate(const char *currLog, const char *saveLog, const struct s
...
             }
         }
+#if defined(WITH_FALLOCATE)
+    } else if (flags & LOG_FLAG_COPYREDUCE) {
+        message(MESS_DEBUG, "reducing %s\n", currLog);
+
+        if (!debug) {
+            if (fdsave >= 0)
+                fsync(fdsave);
+            block_bytes = ((off_t)read_bytes / sb->st_blksize) * sb->st_blksize;
```
cgzones, line 1236: `(off_t)read_bytes` unnecessary cast

config.c
kdudka: `unknown option` sounds clear to me.  I would not over-complicate this code, which is nearly unmaintanable already.

logrotate.c
```diff
@@ -1221,6 +1226,28 @@ static int copyTruncate(const char *currLog, const char *saveLog, const struct s
                 goto fail;
             }
         }
+#if defined(WITH_FALLOCATE)
+    } else if (flags & LOG_FLAG_COPYREDUCE) {
```
kdudka, line 1230: I find it counter-intuitive that, if both `copytruncate` and `copyreduce` are set, then `copytruncate` silently takes precedence.

logrotate.c
```diff
@@ -1221,6 +1226,28 @@ static int copyTruncate(const char *currLog, const char *saveLog, const struct s
...
+            block_bytes = ((off_t)read_bytes / sb->st_blksize) * sb->st_blksize;
+            if (block_bytes > 0) {
+                if (fallocate(fdcurr, FALLOC_FL_COLLAPSE_RANGE, 0L, block_bytes)) {
+                    message(MESS_ERROR, "error reducing %s: %s\n", currLog,
+                            strerror(errno));
+                    goto fail;
+                }
+            }
+            if (fallocate(fdcurr, FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE, 0L, read_bytes - block_bytes)) {
+                message(MESS_ERROR, "error reducing %s: %s\n", currLog,
```
kdudka, line 1245: Could you please make the diagnostic messages more specific, such that it is clear from the verbose output at which step the operation failed?

kdudka (changes requested):
I like this feature.  Thank you for working on it!  Please mention the new configuration directive also in `logrotate.8.in`.  We should probably declare the feature as experimental for now.  Some minor comments inline...

config.c
yosshy: "unsupported option" looks good for me.

logrotate.c
yosshy: ok. I'll add a check to display an error if both copytruncate and copyreduce are specified.

logrotate.c
yosshy: I see.

logrotate.c
yosshy: got it.

logrotate.h
```diff
@@ -39,6 +40,15 @@
 #define WITH_ACL 1
 #endif
 
+#if defined(HAVE_LINUX_FALLOC_H) && defined(HAVE_STRUCT_STAT_ST_BLKSIZE)
+#   include <linux/falloc.h>
+#   define WITH_FALLOCATE 1
+#   if !defined(FALLOC_FL_COLLAPSE_RANGE) || !defined(FALLOC_FL_PUNCH_HOLE) || !defined(FALLOC_FL_KEEP_SIZE)
+#      undef WITH_FALLOCATE
```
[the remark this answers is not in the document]
yosshy: got it.

yosshy (commented):
@cgzones @kdudka Thank you for your review. I'll make a new patchset soon.

logrotate.c
```diff
@@ -1210,6 +1215,12 @@ static int copyTruncate(const char *currLog, const char *saveLog, const struct s
     }
 
     if (flags & LOG_FLAG_COPYTRUNCATE) {
+#if defined(WITH_FALLOCATE)
+        if (flags & LOG_FLAG_COPYREDUCE) {
+            message(MESS_ERROR, "copytruncate and copyreduce are mutually exclusive\n");
+            goto fail;
```
kdudka, line 1221: I propose to make it warning only.  Otherwise logs that use `copytruncate` or `copyreduce` would stop rotating if someone used the other configuration directive on global scope.

kdudka (approved):
Thanks for the update!  One minor remark though.

logrotate.c
yosshy: I see. I removed "goto fail".

yosshy:
Thank you for your review. I'll put another patchset soon.

logrotate.c
```diff
@@ -1210,6 +1225,12 @@ static int copyTruncate(const char *currLog, const char *saveLog, const struct s
     }
 
     if (flags & LOG_FLAG_COPYTRUNCATE) {
+#if defined(WITH_FALLOCATE)
+        if (flags & LOG_FLAG_COPYREDUCE) {
+            message(MESS_NORMAL, "WARNING: copytruncate and copyreduce are mutually exclusive;\n");
+            message(MESS_NORMAL, "copyreduce is disabled.\n");
```
kdudka, line 1231: I see that the patch was reworked.  Now the diagnostic message is misleading again (and maybe the program logic not fully correct).  It basically tells the user that `copyreduce` was ignored after the flag already affected how `sparse_copy()` was executed.  Please put there also a code comment to explain the `skip_null_head` block in `sparse_copy()`.  Thanks!

test/test-0089.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
cgzones, line 1: Please use pure Bourne Shell, see https://github.com/logrotate/logrotate/commit/03110cbb9fdfe63092c4991192c2e2223516cc2b

Forza-tng:
> > This means that loggers using file descriptors should continue to write to the new, swapped-in file.
> 
> @Forza-tng That is not how I read the man page. The operation just swaps the file **names**. It does not say anything about already open file descriptors.

You are right. I should have found some tests before suggesting this.

Could this be interesting? https://lore.kernel.org/linux-fsdevel/161723933214.3149451.12102627412985512284.stgit@magnolia/
It's an ioctl to swap byte ranges between files.

yosshy:
@kdudka @Forza-tng : Thank you for comments.

> Could this be interesting? https://lore.kernel.org/linux-fsdevel/161723933214.3149451.12102627412985512284.stgit@magnolia/ It's an ioctl to swap byte ranges between files.

Interesting. Looks better than FALLOC_FL_COLLAPSE_RANGE and so on.

Forza-tng:
> @kdudka @Forza-tng : Thank you for comments.
> 
> > Could this be interesting? https://lore.kernel.org/linux-fsdevel/161723933214.3149451.12102627412985512284.stgit@magnolia/ It's an ioctl to swap byte ranges between files.
> 
> Interesting. Looks better than FALLOC_FL_COLLAPSE_RANGE and so on.

According to the author of that patch, FIEXCHANGE_RANGE will also work against holes. So you can have an fallocated file (holes) that you swap a range with a normal file.

rk94655:
When can this PR be merged, need this feature as I am facing this issue https://github.com/logrotate/logrotate/issues/363

PR #368, opened 2020-12-18
Ensure the type for configuration flags is big enough

cgzones:
On 32-bit platform int might be only 16 bit wide.
With #361 we would exceed this range.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -18,6 +18,8 @@ AC_STRUCT_ST_BLOCKS
 
 AC_CANONICAL_HOST
 
+AC_TYPE_UINT32_T
+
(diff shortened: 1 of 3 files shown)


kdudka (approved):
Looks good.  I would just write `1U` rather than `1u` to keep the code consistent, and perhaps to avoid unnecessary static analyzer reports in the future: https://rules.sonarsource.com/c/RSPEC-818

kdudka:
@cgzones This change seems to be included in #407, too.  I would prefer to keep it at a single place for review.  Please either close this pull request, or merge it and rebase #407 on top of it.

PR #372, opened 2021-01-26
Lambda user

rmanfredi:
Small changes required to allow a normal user to run logrotate on files that do not belong to user.

This is useful when a container writes logs to a mapped directory, belonging to the user, but whose files have "foreign ids" (those of the container environment).  The ability to rotate the logs and have the logs compressed is important but requires that logs be opened read-only when compressing them, and that no attempt be made to "chown" them to the same uid/gid as the original log file.

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -774,6 +774,8 @@ static int compressLogFile(const char *name, const struct logInfo *log, const st
     int error_printed = 0;
     char *prevCtx;
     pid_t pid;
+    int in_flags;
+    const char *in_how;
 
     message(MESS_DEBUG, "compressing log with: %s\n", log->compress_prog);
     if (debug)
@@ -789,10 +791,18 @@ static int compressLogFile(const char *name, const struct logInfo *log, const st
     compressedName = alloca(strlen(name) + strlen(log->compress_ext) + 2);
     sprintf(compressedName, "%s%s", name, log->compress_ext);
 
-    /* We do not need write permission on that file to compress it */
-    if ((inFile = open(name, O_RDONLY | O_NOFOLLOW)) < 0) {
-        message(MESS_ERROR, "unable to open %s for compression: %s\n",
-            name, strerror(errno));
+    in_flags = O_NOFOLLOW;
+    if (log->flags & LOG_FLAG_SHRED) {
+        /* need write access for shredding */
+        in_flags |= O_RDWR;
+        in_how = "read-write";
+    } else {
+        in_flags |= O_RDONLY;
+        in_how = "read-only";
+    }
+    if ((inFile = open(name, in_flags)) < 0) {
+        message(MESS_ERROR, "unable to open %s (%s) for compression: %s\n",
+            name, in_how, strerror(errno));
         return 1;
     }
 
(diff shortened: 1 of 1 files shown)


logrotate.c
```diff
@@ -785,8 +789,10 @@ static int compressLogFile(const char *name, const struct logInfo *log, const st
     compressedName = alloca(strlen(name) + strlen(log->compress_ext) + 2);
     sprintf(compressedName, "%s%s", name, log->compress_ext);
 
-    if ((inFile = open(name, O_RDWR | O_NOFOLLOW)) < 0) {
```
cgzones, line 788: We need write access in case we want to shred the file after compression.
https://github.com/logrotate/logrotate/blob/0a900b9435522b1314a39ead26ee7cccc91f5674/logrotate.c#L895-L898

So we might only open with `O_RDONLY` in case of `!(log->flags & LOG_FLAG_SHRED)`.

Side note:
Maybe we should add a check in `shred_file()` whether the passed file-descriptor is opened in write mode (via `fcntl(fd, F_GETFL)` and check for `O_WRONLY` or `O_RDWR`)?

logrotate.c
```diff
@@ -582,8 +582,12 @@ static int createOutputFile(const char *fileName, int flags, const struct stat *
         return -1;
     }
 
-    if ((sb_create.st_uid != sb->st_uid || sb_create.st_gid != sb->st_gid) &&
-            fchown(fd, sb->st_uid, sb->st_gid)) {
+    /* Only attempt to set user/group if running as root */
+    if (
+        0 == geteuid() &&
```
cgzones, line 587: So this happens if an unprivileged user rotates a file of another unprivileged user?
Should this be supported by default, as the rotated log files are now having different ownership than the original log file?

kdudka: I see no reason why an unprivileged user should be prevented from rotating files they have access to.  Erroring out in this case could hardly help anybody.

logrotate.c
kdudka: Good point about the shred scenario.  Not sure about adding the extra check.  The file descriptor is under logrotate's control, isn't it?

rmanfredi: Do you want me to add the `(log->flags & LOG_FLAG_SHRED)` to select the proper `open()` flags?  I had missed the `shred` option because I do not use it.

logrotate.c
rmanfredi: Indeed @kdudka.  The directory where I `logrotate` the container logs looks like:
    -rw-r--r-- 1 _apt root       0 Jan 28 01:02 error.log
    -rw-r--r-- 1 _apt root    7912 Jan 27 11:08 error.log.1
    -rw-r--r-- 1 ram  ram       20 Jan 26 01:02 error.log.2.gz
So yes, the log files are given to me, but then I have read-access to the logs and I could copy/rename/compress them myself, since I have write access to the directory containing them...  Anything I can do manually should be possible with logrorate...

cgzones: Convinced.
For consistency the check should be against `ROOT_UID` instead of plain `0`.

logrotate.c
cgzones: Maybe something like:
```c
int flags = O_NOFOLLOW;
/* need write access for shredding */
if (log->flags & LOG_FLAG_SHRED) {
    flags |= O_RDWR;
} else {
    flags |= O_RDONLY;
}
```
?

kdudka (approved):
Thanks for the update!  Looks good to me.

cgzones (approved):

kdudka:
Merging, thanks!

PR #394, opened 2021-06-02
Misc

cgzones:
* Do not use incorrect stat information
  Do not the device number of the configuration file when verifying an olddir configuration after creating the olddir.

  Use a local stat struct in one more place to avoid misuses.

* do_mkdir: minimize race potential
  Perform chown and chmod after directory creation on a file descriptor instead of the path.
  There is still potential for races in between `mkdir(2)` and `open(2)`, but it does not apply to non directories any more.

* Use ROOT_UID instead of 0
  Support non zero root uids.

### DIFF
--- a/config.c
+++ b/config.c
@@ -392,19 +392,31 @@ static char *readAddress(const char *configFile, int lineNum, const char *key,
 
 static int do_mkdir(const char *path, mode_t mode, uid_t uid, gid_t gid) {
     if (mkdir(path, mode) == 0) {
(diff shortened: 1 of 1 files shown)


kdudka (changes requested):
Could we please write `sb_glob_item`, `sb_src_dir`, and `sb_dst_dir` (or something like that) rather than  `sb2,` `sb2`, and `sb3`?

kdudka (approved):

Issue #395, opened 2021-06-03
[Feature Request] allow to disable state management


tpo:
Hi,

I'd be nice if it'd be possible to tell `logrotate` "do not try to update your state file". I can see two ways to do this: add some flag `--do-not-update-state-file` or pass `/dev/null` as a state file. In the latter case, if `logrotate` would see that it got `/dev/null`, then it wouldn't bother renaming the file etc. I think the `/dev/null` variant would be less work.

Rationale: recently I've been often using the following pattern:

    # cat /etc/cron.d/some_menial_task
    0 3 * * www-data root clean_up

    # cat /usr/local/bin/clean_up
    #!/bin/bash
    if cleanup_running; then
      exit 0
    fi
    (
      do_stuff_that_takes_long
      maybe_more_than_a_day
    ) > /var/log/clean_up/log
    logrotate --force /etc/logrotate.d.manual/clean_up

That script runs daily (maybe) and takes long, maybe a day or longer. Once it's done i want to rotate logs. Since I do *not* want to rotate logs daily, but only *after* the `clean_up` has run, I can not put the rotation job into `/etc/logrotate.d/clean_up`, but instead I have to call it manually.

Since the `clean_up` script is running as non-root, I just want `logrotate` to rotate the logs as that user and not bother changing ownerships of the logs. And I don't want it to bother updating it's state file either, since it's me who controls rotation logic and policy, and since the state file belongs to `root` which logrotate can't change as user `www-data`.

In short, instead of writing my own rotation logic I, as non-root, want to use the tool at hand that knows it all and does it best: `logrotate`. So it'd be nice to be able to call `logrotate` and be done with it, without `logrotate` trying to manage its state file.

So I'm putting this proposal up here for consideration.

Thanks a lot for `logrotate`!
*t

kdudka:
Thank you for explaining the situation!  One thing is still unclear to me though: Why do you want to prevent logrotate from writing the state file?  Given the fact that you run `logrotate --force`, contents of the state file is going to be ignored anyway on the next run.  So you could probably use any non-default path as the state file, e.g. `logrotate --force --state /tmp/logrotate.status ...`.  Am I missing anything?

tpo:
Hi @kdudka,

thanks a lot for the kind response!

>  So you could probably use any non-default path as the state file, e.g. logrotate --force --state /tmp/logrotate.status

Yes that's correct I could do that. However that's quite cumbersome and the question is why bother writing the status file in the first place?

[...]

PR #396, opened 2021-06-03
add documentation for --state /dev/null special case

tpo:
This is documentation for the suggested implementation of #395 aka `--state /dev/null` will not write any state file.


kdudka:
Merging, thanks!

PR #397, opened 2021-06-07
RFC: forbid rotating hardlinks as root

cgzones:
Related: #237 

To avoid any kind of attack scenarios with malicious hard-links to privileged file (like */etc/shadow*) do not rotate hard-links as root (we already do not rotate symbolic-links).

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -225,6 +225,16 @@ static int switch_user_back_permanently(void) {
     return ret;
 }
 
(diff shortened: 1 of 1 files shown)


kdudka:
Thanks for the proposal!  Could you please send me privately the attack scenario that this change would prevent?

kdudka:
I believe the proposed changes are now covered by #407, #400, #398, #399, and #401.  Could we please close this pull request?

PR #399, opened 2021-06-11
Do not open symbolic link log files by accident

cgzones:
We do check a priori if a log file is a symbolic link and skip such
files.  Avoid race conditions where a regular file gets replaced by a
symbolic link after that check, causing logrotate to rotate the content
of the link target.

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -225,6 +225,28 @@ static int switch_user_back_permanently(void) {
     return ret;
 }
 
+static int open_logfile(const char *path, int write_access) {
+    int fd;
+    struct stat sb;
+
+    fd = open(path, O_NOFOLLOW | (write_access ? O_RDWR : O_RDONLY));
+    if (fd < 0)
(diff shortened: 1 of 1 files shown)


logrotate.c
```diff
@@ -225,6 +225,33 @@ static int switch_user_back_permanently(void) {
     return ret;
 }
 
+static int open_logfile(const char *path, int flags) {
+    int fd;
+    struct stat sb;
+
+    if ((flags & (O_RDONLY | O_RDWR)) != flags) {
```
kdudka, line 232: I do not understand this check.  `flags` is under logrotate's control, isn't it?  Is it meant to be `assert(...)` instead?  Please put there a comment in any case.

cgzones: Refactored the interface of `open_logfile` by not passing any flags at all, just a boolean whether to open with write access.

kdudka (approved):

PR #400, opened 2021-06-11
Add defence in depth when dropping privileges

cgzones:


logrotate.c
```diff
@@ -203,12 +203,18 @@ static int switch_user_permanently(const struct logInfo *log) {
...
     }
 
-    if (user != ROOT_UID && setuid(ROOT_UID) != -1) {
+    if (user != ROOT_UID && (setuid(ROOT_UID) != -1 || seteuid(ROOT_UID) != -1)) {
         message(MESS_ERROR, "failed to switch user permanently, able to switch back (pid %d)\n",
                 getpid());
         return 1;
     }
 
+    if (chdir("/") != 0) {
```
kdudka, line 212: Why are we changing the working directory?  Is it an observable change of behavior?  If yes, is it backward-compatible?  I think this should be documented in any case.

cgzones: I think currently logrotate does not give any guarantee about the current working directory (when spawning the compress or mail process or when running any scripts).
In config.c we call `chdir()`, but be at the end we return into the previously directory.
When run by systemd the working directory is by default set to `/`, see https://www.freedesktop.org/software/systemd/man/systemd.exec.html section `WorkingDirectory=`.
I am not sure if cron implementations (cronie, ...) give any guarantees.
Adding to the man-page that the current working directory in scripts is unspecified and should not be relied on would make sense to me.

kdudka: Thank you for digging the information.  So the change should not be needed when logrotate is triggered by systemd as a system service, which is what many Linux distros do these days.

cgzones: It makes a difference when logrotate is started from a root crontab, where the working directory is the home directory, i.e. `/root`.

kdudka: I have no strong opinion on this.  It can improve security in some corner cases but also break compatibility on other corner cases.  We should probably prefer the former to stay on the safe side.

kdudka (approved):

Issue #405, opened 2021-06-21
Travis CI


cgzones:
Since June 15th travis-ci.org is down and it has been replaced by travis-ci.com.
With this transition the quota/credit system has been changed with quite some limitations.
logrotate should fit their category of open-source projects, but there still might be obstacles, see for example https://daniel.haxx.se/blog/2021/06/14/bye-bye-travis-ci/.

Option I is to move to travis-ci.com and see how the services is continuing.
Option II is to move to a new CI service, e.g. GitHub Actions.

kdudka:
I am fine with both.  We definitely need some CI that works.

cgzones:
I tried to activate logrotate on travis-ci.com, but (maybe because I lack some permissions) it did not work.
Can you try to activate or move from travis-ci.org?

kdudka:
Done.  I was able to trigger a build: https://www.travis-ci.com/github/logrotate/logrotate/builds/230427318

I chose the Free plan, which consists of 10000 credits, which is going to be spent quickly I am afraid.  Then we will probably need to apply for some OSS approval to get additional credits.

kdudka:
Closing per https://github.com/logrotate/logrotate/pull/406#issuecomment-1084625776

PR #406, opened 2021-06-21
Add GitHub Actions CI configuration

cgzones:
Covers compilers from Ubuntu 20.20 and Clang analysis and sanitizer jobs.
Drop the respective jobs from travis to reduce the load on them.


kdudka (approved):

cgzones:
Rebased on top of #409.

cgzones:
The builds on travis-ci.com have stopped due to a negative credit balance.
Should we bother requesting `OSS only credits` from their support or move completely to GitHub actions?

cgzones:
The only downside is we are losing the build coverage of GCC versions prior version 7.

kdudka:
Looks good but I cannot see any CI jobs associated with this pull request.  Do we yet need to enabled them somewhere?

> The only downside is we are losing the build coverage of GCC versions prior version 7.

That sounds acceptable to me.  The last supported release of GCC by upstream seems to be 9.4.

cgzones:
GitHub Actions CI ran and succeeded on #430.
> Do we yet need to enable them somewhere?

I don't think so, probably just to avoid random pull requests mining crypto in added CI jobs.

fabiang:
I think this PR fixed #405 :)

PR #407, opened 2021-06-22
Do not rotate hard links by default

cgzones:
Hard links are quite unusual to be a target of log rotation.
They can be subject of attacks since users might create hard links to
privileged files, like /etc/shadow, leading programs running as root to
modify the original privileged file through this link.

Add configuration directives `hardlink` and `nohardlink` to control
whether to rotate had links, whereby `nohardlink` is the default.

logrotate already does not rotate symbolic links.

Alternative to #397.

### DIFF
--- a/config.c
+++ b/config.c
@@ -1126,6 +1126,10 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
                         newlog->flags |= LOG_FLAG_SHRED;
                     } else if (!strcmp(key, "noshred")) {
                         newlog->flags &= ~LOG_FLAG_SHRED;
+                    } else if (!strcmp(key, "allowhardlink")) {
+                        newlog->flags |= LOG_FLAG_ALLOWHARDLINK;
+                    } else if (!strcmp(key, "noallowhardlink")) {
+                        newlog->flags &= ~LOG_FLAG_ALLOWHARDLINK;
                     } else if (!strcmp(key, "sharedscripts")) {
                         newlog->flags |= LOG_FLAG_SHAREDSCRIPTS;
                     } else if (!strcmp(key, "nosharedscripts")) {
--- a/logrotate.8.in
+++ b/logrotate.8.in
@@ -467,6 +467,16 @@ Do not use \fBshred\fR when deleting old log files.  See also \fBshred\fR.
 Asks GNU \fBshred\fR(1) to overwrite log files \fBcount\fR times before
 deletion.  Without this option, \fBshred\fR's default will be used.
 
+.TP
+\fBallowhardlink\fR
+Rotate files with multiple hard links; this is off by default.  The target file
+might get emptied, e.g. with \fBshred\fR or \fBcopytruncate\fR.  Use with
+caution, especially when the log files are rotated as root.
+
+.TP
+\fBnoallowhardlink\fR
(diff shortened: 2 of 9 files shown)


kdudka:
Thanks!  I like the idea.  It makes it easy to restore the old behavior globally in case it surprises anybody after update.  I would only reword the documentation such that it better reflects what the code actually does.  It does not a priory refuse to rotate any log files with multiple hard links, does it?  The option only takes an effect if `shred` or `copytruncate` is triggered, as I understand it.  And we should also refer to these cases as files with *multiple* hard links.  According to `stat(2)` man page, `st_nlink` is a number of hard links.  So all the files we usually rotate (with `st_nlink == 1`) have a hard link by definition.

cgzones:
I'll update the wording with regard to *file with multiple hard links*.

Currently this implementation skips files with multiple hard links by default **regardless** of other directives (`shred`, `coptruncate`).
Are hard-linked log files even a thing? Most files with multiple hard links are binaries or shared libraries, like `/usr/bin/uncompress` - `/usr/bin/gunzip`.

kdudka:
My bad.  `test-0090` actually does not use any of those options.  Please embed also the following patch:
```diff
--- a/logrotate.c
+++ b/logrotate.c
@@ -400,38 +400,39 @@ static int setSecCtx(int fdSrc, const char *src, char **pPrevCtx)
 static int setSecCtxByName(const char *src, const struct logInfo *log, char **pPrevCtx)
 {
     int hasErrors = 0;
 #ifdef WITH_SELINUX
     int fd;

     if (!selinux_enabled)
         /* pretend success */
         return 0;

     fd = open_logfile(src, log, 0);
     if (fd < 0) {
         message(MESS_ERROR, "error opening %s: %s\n", src, strerror(errno));
         return 1;
     }
     hasErrors = setSecCtx(fd, src, pPrevCtx);
     close(fd);
 #else
     (void) src;
+    (void) log;
     (void) pPrevCtx;
 #endif
     return hasErrors;
 }

 static void restoreSecCtx(char **pPrevCtx)
 {
 #ifdef WITH_SELINUX
     if (!*pPrevCtx)
         /* no security context saved for restoration */
         return;

     /* set default security context to the previously stored one */
     if (selinux_enabled && setfscreatecon_raw(*pPrevCtx) < 0)
         message(MESS_ERROR, "setting default context to %s: %s\n", *pPrevCtx,
                 strerror(errno));

     /* free the memory allocated to save the security context */
     freecon(*pPrevCtx);
```
... to eliminate the following warning:
```
logrotate.c: In function 'setSecCtxByName':
logrotate.c:400:67: warning: unused parameter 'log' [-Wunused-parameter]
  400 | static int setSecCtxByName(const char *src, const struct logInfo *log, char **pPrevCtx)
      |                                             ~~~~~~~~~~~~~~~~~~~~~~^~~
```
... when compiled without SELinux.

cgzones:
Updated the wording and included the fix for the compiler warning without SELinux support (thanks for spotting).

What's your take on the default setting:
* disable completely by default (currently done by this pr)
* disable only when used with `copytruncate` or `shred`

kdudka:
Thanks for the update!  I think that having `allowhardlink` disabled by default is fine.  If it breaks something, it is easy to restore the old behavior either globally or per log file.

kdudka (approved):

Issue #410, opened 2021-07-26
"keyword 'size' not properly separated, found 0x3d" warning


GimmeHardware:
The man page states values are separated from directives by whitespace and/or an optional =.

But this change:
[https://github.com/logrotate/logrotate/commit/2b588b5ec2e5c27bee857c4abeddafa6a9602ebc](https://github.com/logrotate/logrotate/commit/2b588b5ec2e5c27bee857c4abeddafa6a9602ebc)
does not accept the optional =.

For example "size=+2048k" in the mcelog logrotate file works with logrotate 3.18.0 but triggers this warning with 3.18.1.
[https://git.kernel.org/pub/scm/utils/cpu/mce/mcelog.git/tree/mcelog.logrotate](https://git.kernel.org/pub/scm/utils/cpu/mce/mcelog.git/tree/mcelog.logrotate)
[https://www.linuxquestions.org/questions/showthread.php?s=cb29de3ab1aa6aa6407d395d3c9d07d4&p=6253331#post6253331](https://www.linuxquestions.org/questions/showthread.php?s=cb29de3ab1aa6aa6407d395d3c9d07d4&p=6253331#post6253331)

cgzones:
Thanks for reporting.
#411 should silence this unintended warning.

PR #411, opened 2021-07-26
Do not warn on key value pair separated by only an equal sign

cgzones:
Do not warn if a configuration directive is specified with the key and
value separated by just an equal sign, like:

    size=+2048k

The warning is intended for the usage of:

    size2048k

Fixes: 2b588b5e ("Log if keyword is not properly separated")
Fixes: #410


kdudka:
Looks good.  Thanks!

kdudka (approved):

PR #414, opened 2021-08-05
Assert valid fd in shred_file when shredding

cgzones:
Coverity gets confused when shred_file() is called from removeLogFile()
with LOG_FLAG_SHRED not set.  In this case shred_file() is passed an
invalid file descriptor but it gets never used.
Add an assert statement to silence this false-positive.

    *** CID 372221:  Integer handling issues  (NEGATIVE_RETURNS)
    /logrotate.c: 779 in removeLogFile()
    773                 message(MESS_ERROR, "error opening %s: %s\n",
    774                         name, strerror(errno));
    775                 return 1;
    776             }
    777         }
    778
    >>>     CID 372221:  Integer handling issues  (NEGATIVE_RETURNS)
    >>>     "fd" is passed to a parameter that cannot be negative.
    779         if (!debug && shred_file(fd, name, log)) {
    780             message(MESS_ERROR, "Failed to remove old log %s: %s\n",
    781                     name, strerror(errno));
    782             result = 1;
    783         }
    784

Also assert that the file descriptor is eligible for modifying.

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -1,4 +1,5 @@
 #include "queue.h"
+#include <assert.h>
 #include <limits.h>
 #include <ctype.h>
 #include <dirent.h>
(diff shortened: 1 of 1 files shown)


kdudka (approved):
I cannot see such a confusion with up2date version of Coverity but the change seems harmless.

cgzones:
> I cannot see such a confusion with up2date version of Coverity but the change seems harmless.

I wonder if it's then better to not add the assert statement to not influence the analysis, in particular with respect to possible logic issues of future changes.

kdudka:
It is up to you.

cgzones:
Keep it as it is for now.

PR #415, opened 2021-08-07
Fix full_write on incomplete write

cgzones:
Currently on an incomplete write(3) the subsequent write(3) will again
use the same starting position in the given buffer, which was already
successfully written, leading to data corruption.

Found by clang-13

    logrotate.c:1140:17: warning: variable 'ptr' set but not used [-Wunused-but-set-variable]
    const char *ptr = (const char *) buf;
                ^

Fixes: f1dc0d9 ("Allow rotation of sparse files with copytruncate")

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -1137,18 +1137,18 @@ static int is_nul (void const *buf, size_t bufsize)
 static size_t full_write(int fd, const void *buf, size_t count)
 {
     size_t total = 0;
-    const char *ptr = (const char *) buf;
+    const unsigned char *ptr = (const unsigned char *) buf;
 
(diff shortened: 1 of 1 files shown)


kdudka:
The actual fix looks good.  I am not sure about the added `unsigned` -- does it help in any way?

Another preexisting issue with the code is that it should not check the value of `errno` unless `write()` returns `-1`.

cgzones:
> Another preexisting issue with the code is that it should not check the value of errno unless write() returns -1.

True, I updated the patch.

kdudka (approved):
Looks good.  Thanks for the update!

Issue #425, opened 2021-10-13
Why is the logrotate.status file missing?


matthuo333:
When I run debug to test my configuration file, I get the following error:

reading config file /etc/logrotate.d/smartnode
Reading state from file: /var/lib/logrotate.status
**error: error opening state file /var/lib/logrotate.status: No such file or directory**
Allocating hash table for state file, size 64 entries

Handling 1 logs

rotating pattern: /var/log/smartnode.log after 1 days (7 rotations)
empty log files are not rotated, only log files >= 1048576 bytes are rotated, old logs are removed
considering log /var/log/smartnode.log
Creating new state
  Now: 2021-10-13 12:52
  Last rotated at 2021-10-13 12:00
  log does not need rotating (log has already been rotated)

The current OS is OpenWrt.

PR #426, opened 2021-10-13
Avoid confusing error message in debug mode

cgzones:
In debug mode the state file is not created if not already existent in
lockState().  Do not print the message

    error: error opening state file /var/lib/logrotate.status: No such file or directory

while trying to read the state file, since in debug mode this is not an
error.

Related: #425

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -2789,16 +2789,19 @@ static int readState(const char *stateFilename)
         /* treat non-openable file as an empty file for allocateHash() */
         f_stat.st_size = 0;
 
-        /* the statefile should exist, lockState() already created an empty
-         * state file in case it did not exist initially */
-        message(MESS_ERROR, "error opening state file %s: %s\n",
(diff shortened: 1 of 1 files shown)


logrotate.c
```diff
@@ -2789,16 +2789,21 @@ static int readState(const char *stateFilename)
...
-                stateFilename, strerror(errno));
-
         /* Do not return until the hash table is allocated.
          * In debug mode the state file might not exist,
          * cause lockState() is not called */
         if (!debug) {
+            message(MESS_ERROR, "error opening state file %s: %s\n",
+                    stateFilename, strerror(errno));
             rc = 1;
+        } else {
```
kdudka, line 2799: Could you please write it as a linear `if ... else if ... else ...` chain without the unnecessary level of nesting?

cgzones: changed

kdudka: Thanks for the update!

PR #427, opened 2021-10-21
config.c: enforce stricter parsing of config files

felixwilhelm:
A number of recent exploits have abused logrotate's lax parsing of config files as an easy way to turn privileged file write bugs into code execution:
- https://alephsecurity.com/2021/10/20/sudump/
- https://flattsecurity.medium.com/cve-2020-15702-race-condition-vulnerability-in-handling-of-pid-by-apport-4047f2e00a67#e688
- https://alephsecurity.com/2021/02/16/apport-lpe/

These exploits depend on bugs in other parts of the OS (apport, kernel, ..) to write partially controlled files to the logrotate config directory. Targeting logrotate is highly useful for attackers, because it will ignore invalid parts of config files and continue parsing until it finds a valid configuration.
This behaviour makes exploitation of tricky bugs such as the linked coredump vulnerabilities much easier.

This PR enforces stricter parsing of config files, by turning some of the already existing parser warnings into hard errors. It does not have any effect on users using valid configuration files, but might be a breaking change for people relying on the current handling of invalid files. As current versions already print an error message when encountering such files, no widely used configurations should be affected. 

Overall, I think this would be a nice defense-in-depth change to improve the security of many downstream logrotate users.

### DIFF
--- a/config.c
+++ b/config.c
@@ -1107,12 +1107,13 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
                     if (key == NULL) {
                         message(MESS_ERROR, "%s:%d failed to parse keyword\n",
                                 configFile, lineNum);
-                        continue;
+                        RAISE_ERROR();
                     }
                     if (!isspace((unsigned char)*start) && *start != '=') {
-                        message(MESS_NORMAL, "%s:%d keyword '%s' not properly"
+                        message(MESS_ERROR, "%s:%d keyword '%s' not properly"
                                 " separated, found %#x\n",
                                 configFile, lineNum, key, *start);
+                        RAISE_ERROR();
                     }
                     if (!strcmp(key, "compress")) {
                         newlog->flags |= LOG_FLAG_COMPRESS;
@@ -2004,7 +2005,7 @@ duperror:
                     message(MESS_ERROR, "%s:%d lines must begin with a keyword "
                             "or a filename (possibly in double quotes)\n",
                             configFile, lineNum);
-                    state = STATE_SKIP_LINE;
+                    RAISE_ERROR();
                 }
                 break;
             case STATE_SKIP_LINE:
--- a/test/test-0102.sh
+++ b/test/test-0102.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+. ./test-common.sh
+
+cleanup 102
+
+# ------------------------------- Test 102 ------------------------------------
+# test invalid config file with binary content
+preptest test.log 102 1
+
+$RLR test-config.102 --force
+
+if [ $? -eq 0 ]; then
+   echo "No error, but there should be one."
+   exit 3
+fi
--- a/test/test-0103.sh
+++ b/test/test-0103.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+. ./test-common.sh
+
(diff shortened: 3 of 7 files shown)


felixwilhelm:
@kdudka friendly ping.

kdudka:
Sorry for the delay!  I was on vacation last week.  The proposed changes look fair to me.  I would appreciate an additional review by @cgzones though.  He is the original author of the `keyword ... not properly separated` diagnostic message (and the most active contributor of logrotate these days).

config.c
```diff
@@ -2004,7 +2005,7 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
                     message(MESS_ERROR, "%s:%d lines must begin with a keyword "
                             "or a filename (possibly in double quotes)\n",
                             configFile, lineNum);
-                    state = STATE_SKIP_LINE;
+                    RAISE_ERROR();
```
cgzones, line 2008: I think in this case we want to abort reading the whole file and not parse any potential following blocks, so `goto error;` would be more appropriate.

test/test-0102.sh
```diff
@@ -0,0 +1,16 @@
...
+
+cleanup 102
+
+# ------------------------------- Test 102 ------------------------------------
+# test invalid config file with binary content
+preptest test.log 102 1
+
+$RLR test-config.102 --force
+
+if [ $? -eq 0 ]; then
```
cgzones, line 13: Please also check that `test.log` has not been rotated.

test/test-0103.sh
```diff
@@ -0,0 +1,16 @@
...
+
+cleanup 103
+
+# ------------------------------- Test 103 ------------------------------------
+# test invalid config file with unknown keywords
+preptest test.log 103 1
+
+$RLR test-config.103 --force
+
+if [ $? -eq 0 ]; then
```
cgzones, line 13: Please also check that `test.log` has not been rotated.

cgzones (changes requested):
Please add test cases for the following configurations:

```
&DIR&/test1.log {
    newkeyword
}
&DIR&/test2.log {
    rotate 1
}
```
and check that test1.log has not been rotated, but test2.log.

```
&DIR&/test1.log {
    g@rbag€[]+#*
}
&DIR&/test2.log {
    rotate 1
}
```
and check that both test1.log and test2.log have not been rotated.

kdudka:
@cgzones My proposal was actually not to treat unknown configuration directives as errors that would prevent log files from being rotated.  logrotate configuration files are often updated independently of logrotate itself and their compatibility with older logrotate releases is important in my opinion.

config.c
cgzones: > My proposal was actually not to treat unknown configuration directives as errors that would prevent log files from being rotated. logrotate configuration files are often updated independently of logrotate itself and their compatibility with older logrotate releases is important in my opinion.

true, so my comment was void.

cgzones:
@kdudka since you approved #431, can I merge this one here and afterwards apply the test suite changes?

kdudka:
Definitely.  Thank you for taking care of it!

cgzones:
Applied, thanks for the contribution.

felixwilhelm:
Thanks for pushing this through! 👍

PR #429, opened 2021-11-12
Allow log rotation across filesystems

kanavin:
The logrotate can't save the log across the different filesystems since
it uses the "rename(const char *oldpath, const char *newpath)" to save
the file.

This patch changes the code to act as the "mv" command (first rename, if failed,
read and write).

Signed-off-by: Robert Yang <liezhi.yang@windriver.com>
Signed-off-by: Alexander Kanavin <alex@linutronix.de>

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -1543,6 +1543,53 @@ static int findNeedRotating(const struct logInfo *log, unsigned logNum, int forc
     return 0;
 }
 
+/* Act as the "mv" command, if rename failed, then read the old file and
+ * write to new file. The function which invokes the mvFile will use
+ * the strerror(errorno) to handle the error message, so we don't have
+ * to print the error message here */
+
+int mvFile (char *oldName, char *newName, struct logInfo *log, acl_type acl)
+{
+    struct stat sbprev;
+    int fd_old, fd_new, n;
+    char buf[BUFSIZ];
+
+    /* Do the rename first */
+    if (!rename(oldName, newName))
+        return 0;
+
+    /* If the errno is EXDEV, then read old file, write newfile and
+     * remove the oldfile */
+    if (errno == EXDEV) {
+        /* Open the old file to read */
+        if ((fd_old = open(oldName, O_RDONLY)) < 0)
+            return 1;
+
+        /* Create the file to write, keep the same attribute as the old file */
+        if (stat(oldName, &sbprev))
+            return 1;
+        else {
+            if ((fd_new = createOutputFile(newName,
+                O_WRONLY | O_CREAT | O_TRUNC, &sbprev, acl, 0)) < 0 )
+                return 1;
+        }
+
+        /* Read and write */
+        while ((n = read(fd_old, buf, BUFSIZ)) > 0)
+            if (write(fd_new, buf, n) != n)
+                return 1;
+
+        if ((close(fd_old) < 0) ||
+            removeLogFile(oldName, log) ||
+            (close(fd_new) < 0))
+            return 1;
+
+        return 0;
+    }
+
+    return 1;
+}
+
 /* find the rotated file with the highest index */
 static int findLastRotated(const struct logNames *rotNames,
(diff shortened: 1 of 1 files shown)


logrotate.c
```diff
@@ -1543,6 +1543,53 @@ static int findNeedRotating(const struct logInfo *log, unsigned logNum, int forc
     return 0;
 }
 
+/* Act as the "mv" command, if rename failed, then read the old file and
+ * write to new file. The function which invokes the mvFile will use
+ * the strerror(errorno) to handle the error message, so we don't have
+ * to print the error message here */
+
+int mvFile (char *oldName, char *newName, struct logInfo *log, acl_type acl)
```
cgzones, line 1551: Please `const` pointers, as this function should not modify any of its data.

logrotate.c
```diff
@@ -1543,6 +1543,53 @@ static int findNeedRotating(const struct logInfo *log, unsigned logNum, int forc
...
+    int fd_old, fd_new, n;
+    char buf[BUFSIZ];
+
+    /* Do the rename first */
+    if (!rename(oldName, newName))
+        return 0;
+
+    /* If the errno is EXDEV, then read old file, write newfile and
+     * remove the oldfile */
+    if (errno == EXDEV) {
```
cgzones, line 1563: Use a early exit in case of `errno != EXDEV`, to avoid the following block and save an indentation level.

logrotate.c
```diff
@@ -1543,6 +1543,53 @@ static int findNeedRotating(const struct logInfo *log, unsigned logNum, int forc
...
+        /* Open the old file to read */
+        if ((fd_old = open(oldName, O_RDONLY)) < 0)
```
cgzones, line 1565: This should use the wrapper `open_logfile(oldName, log, 0)`

logrotate.c
```diff
@@ -1543,6 +1543,53 @@ static int findNeedRotating(const struct logInfo *log, unsigned logNum, int forc
...
+            return 1;
+
+        /* Create the file to write, keep the same attribute as the old file */
+        if (stat(oldName, &sbprev))
+            return 1;
```
cgzones, line 1570: `fd_old` is not closed (also not in the subsequent error branches)

logrotate.c
```diff
@@ -1543,6 +1543,53 @@ static int findNeedRotating(const struct logInfo *log, unsigned logNum, int forc
...
+        else {
+            if ((fd_new = createOutputFile(newName,
+                O_WRONLY | O_CREAT | O_TRUNC, &sbprev, acl, 0)) < 0 )
+                return 1;
+        }
+
+        /* Read and write */
+        while ((n = read(fd_old, buf, BUFSIZ)) > 0)
```
cgzones, line 1578: handle `EINTR`?

logrotate.c
```diff
@@ -1543,6 +1543,53 @@ static int findNeedRotating(const struct logInfo *log, unsigned logNum, int forc
...
+            if (write(fd_new, buf, n) != n)
```
cgzones, line 1579: use `full_write()`

logrotate.c
```diff
@@ -1543,6 +1543,53 @@ static int findNeedRotating(const struct logInfo *log, unsigned logNum, int forc
...
+                return 1;
+
+        if ((close(fd_old) < 0) ||
+            removeLogFile(oldName, log) ||
+            (close(fd_new) < 0))
+            return 1;
```
cgzones, line 1585: if any of the first calls fail, `fd_new` does not get closed

cgzones (changes requested):
There should probably be a unit test, e.g. testing on `/run` (skip test if not root).
Also the SELinux context is not preserved.

kanavin:
Thank you for the in-depth review. I am not the author; I only took the patch and submitted it here to reduce our (Yocto's) mass of custom patches. I have just asked the author to make those changes, so they will come back with a reworked patch in a separate MR - and this can be closed.

PR #431, opened 2021-12-13
config.c: enforce stricter parsing of config files [no. 2]

cgzones:
Same code as #427, plus some more test cases.


kdudka (approved):
Nice!  Thank you for moving it forward!

PR #438, opened 2022-02-15
Support unsigned time_t

willmmiles:
On platforms with unsigned time_t simple subtraction may produce a large positive result instead of the expected negative value.  Use difftime() instead of subtracting to ensure meaningful comparisons.

I ran in to this issue using logrotate on QNX 6.6.  It may be the only POSIX-compliant platform in any use today with an unsigned time_t, so this fix may not be worth the churn, but I thought I'd offer it anyhow.

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -1338,10 +1338,10 @@ static time_t mktimeFromDateOnly(const struct tm *src)
 }
 
 /* return by how many days the date was advanced but ignore exact time */
-static time_t daysElapsed(const struct tm *now, const struct tm *last)
+static long daysElapsed(const struct tm *now, const struct tm *last)
 {
     const double diff = difftime(mktimeFromDateOnly(now),mktimeFromDateOnly(last));
-    return (time_t) (diff / (24 * 3600));
+    return (long) (diff / (24 * 3600));
 }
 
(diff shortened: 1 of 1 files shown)


kdudka:
No objections to use `difftime()` where appropriate but why are you replacing `time_t` with `int`.  On `x86_64` these types have different width (64bit vs. 32bit) and this seems like a step backwards.

willmmiles:
For daysElapsed()?   Because POSIX time_t may be unsigned (as it is on my platform) so it cannot be used when the return value could be negative.   For a result in days, it didn't seem to me like it was worth explicitly overriding the system default width -- a 32-bit day count spans a range of about 5.8 million years.   If that's not enough I can resubmit with int64_t.

kdudka:
We do not seem to use `int64_t` in logrotate yet, so I am not sure whether it is available on all the supported platforms.  But we should at least use `long` rather than `int` to avoid narrowing of the type in the most common case.

willmmiles:
Updated.  I've broken the changes down in to two commits (one for difftime, one for daysElapsed) for clarity.

kdudka:
Merging, thanks for the contribution!

PR #444, opened 2022-04-28
Mention if copying will be skipped

cgzones:
E.g. if rotate count is 0.


kdudka (approved):
Nice.  Thanks for the fixup!

PR #446, opened 2022-05-25
drop world-readable permission on state file

kdudka:
... even when ACLs are enabled.  This is a follow-up to the fix
of CVE-2022-1348.  It has no impact on security but makes the state
file locking work again in more cases.


lzaoral:
LGTM, @kdudka.

PR #448, opened 2022-05-27
lockState: do not print `error:` when exit code is unaffected

kdudka:


cgzones (approved):

kdudka:
Thanks for review!  Merging...

PR #449, opened 2022-05-27
log: unify handling of log levels

kdudka:
Use `MESS_WARN` instead of `MESS_NORMAL` and make it always use the `warning:` prefix.  Also drop `MESS_VERBOSE`, which was not set anywhere.

### DIFF
--- a/config.c
+++ b/config.c
@@ -643,7 +643,7 @@ static void set_criterium(enum criterium *pDst, enum criterium src, int *pSet)
 {
     if (*pSet && (*pDst != src)) {
         /* we are overriding a previously set criterium */
-        message(MESS_VERBOSE, "warning: '%s' overrides previously specified '%s'\n",
+        message(MESS_DEBUG, "note: '%s' overrides previously specified '%s'\n",
                 crit_to_string(src), crit_to_string(*pDst));
     }
     *pDst = src;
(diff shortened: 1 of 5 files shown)


log.c
```diff
@@ -78,8 +81,7 @@ void message(int level, const char *format, ...)
                 priority |= LOG_DEBUG;
                 break;
             case MESS_DEBUG:
-            case MESS_VERBOSE:
-            case MESS_NORMAL:
+            case MESS_WARN:
```
cgzones, line 84: Should `MESS_WARN` maybe translate to a syslog priority of `LOG_WARNING` (or `LOG_NOTICE`)?

kdudka: Yes, that makes perfect sense.

config.c
```diff
@@ -643,7 +643,7 @@ static void set_criterium(enum criterium *pDst, enum criterium src, int *pSet)
 {
     if (*pSet && (*pDst != src)) {
         /* we are overriding a previously set criterium */
-        message(MESS_VERBOSE, "warning: '%s' overrides previously specified '%s'\n",
+        message(MESS_DEBUG, "warning: '%s' overrides previously specified '%s'\n",
```
cgzones, line 646: Since `warning: ` is now a common prefix, maybe rename this one to something similar to `note: `?

kdudka: Good idea!  Renamed and force pushed...

cgzones (approved):

kdudka:
Yes, good point, except that this pull request does not add the `debug:` prefix, which is probably fine.

PR #454, opened 2022-07-25
Support home dir expansion in olddir

cgzones:
Support expanding paths leading with ~/ in olddir, similar to include.

Closes: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1015964


kdudka (approved):
Looks good.  Thanks!

PR #479, opened 2022-12-05
Timestamp updating improvements

cgzones:
* test: check atime and mtime are set after compression

* clarify conditional usage of utimensat

  Commit https://github.com/logrotate/logrotate/commit/0d805cee34d3a9b1937130fb7c7fcf75d940f51f ("logrotate: do not use utimensat() on Darwin") disabled
the usage of utimensat(3) on Darwin.  The actual issue was not the libc
call of utimensat(3), but the more precise struct `stat` members `st_atim`
and `st_mtim`.  Check for those struct members at configure time and
avoid using a general conditional check against `__APPLE__`.

* update file timestamps via descriptor if possible

  Update the atime and mtime of newly created compressed rotated log files
via the race free function futimens(3) (if available).  Also use the
flag AT_SYMLINK_NOFOLLOW to not dereference symlinks via utimensat(3).

  This should prevent theoretical race conditions with regard to the underlying file.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -168,6 +168,7 @@ AC_DEFINE_UNQUOTED([ROOT_UID], [0], [Root user-id.])
 AC_SUBST(ROOT_UID)
 
 AC_CHECK_FUNCS([asprintf madvise secure_getenv strndup utimensat vsyslog])
+AC_CHECK_MEMBERS([struct stat.st_atim, struct stat.st_mtim])
 AC_CONFIG_HEADERS([config.h])
 
 AM_CFLAGS="\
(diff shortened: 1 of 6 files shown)


kdudka:
I have opened pull request #480 to make the CI green again.

@cgzones Initially I tried to squeez in the commit into this pull request but I got `permission denied` error while I was trying to rewrite your `amtime_test` branch.

kdudka:
@cgzones Could you please rebase this pull request on the current master branch to see whether it fixes the CI failure?

kdudka (approved):
Looks good to me.  Thanks!

PR #494, opened 2023-02-21
Clarify debug messages for "rotate -1" + "maxage 0"

cryptogopher:


cryptogopher:
This is invalid, correcting and will post another one

logrotate.c
```diff
@@ -2348,11 +2348,22 @@ static int rotateLogSet(const struct logInfo *log, int force)
...
-    if (log->rotateCount > 0)
-        message(MESS_DEBUG, "(%d rotations)\n", log->rotateCount);
-    else if (log->rotateCount == 0)
+    if (log->rotateCount == 0)
         message(MESS_DEBUG, "(no old logs will be kept)\n");
-
+    else {
+        if (log->rotateCount > 0)
+            message(MESS_DEBUG, "(%d rotations)", log->rotateCount);
+        else if (log->rotateCount == -1)
```
kdudka, line 2356: This condition is redundant, `rotateCount` should never be lower than -1: https://github.com/logrotate/logrotate/blob/0cc114f42c03b7931bd3258405ea13005bfd058d/config.c#L1375

logrotate.c
```diff
@@ -2348,11 +2348,22 @@ static int rotateLogSet(const struct logInfo *log, int force)
...
+            message(MESS_DEBUG, "(unlimited rotations)");
+        
+        if (log->rotateAge >= 0)
+            message(MESS_DEBUG, ", old logs are removed after %d days", log->rotateAge);
+        else if (log->rotateAge == -1) && (log->rotateCount == -1)
```
kdudka, line 2361: Apart from the missing outer parenthesis, the condition can never be true because `rotateAge` can never be lower than zero: https://github.com/logrotate/logrotate/blob/0cc114f42c03b7931bd3258405ea13005bfd058d/config.c#L1415

kdudka (changes requested):

PR #495, opened 2023-02-21
Clarify debug messages for rotate/maxage

cryptogopher:


logrotate.c
```diff
@@ -2348,11 +2348,17 @@ static int rotateLogSet(const struct logInfo *log, int force)
...
         message(MESS_DEBUG, "(no old logs will be kept)\n");
-
+    else if (log->rotateCount > 0) && (log->rotateAge > 0)
+        message(MESS_DEBUG, "(%d rotations), old logs are removed after %d days\n", log->rotateCount, log->rotateAge);
+    else if (log->rotateCount > 0) && (log->rotateAge == 0)
+        message(MESS_DEBUG, "old logs are removed\n");
+    else if (log->rotateCount == -1) && (log->rotateAge > 0)
+        message(MESS_DEBUG, "(unlimited rotations), old logs are removed after %d days\n", log->rotateAge);
+    else if (log->rotateCount == -1) && (log->rotateAge == 0)
+        message(MESS_DEBUG, "old logs are kept forever\n");
```
kdudka, line 2360: I would deduplicate the conditions like this:
```c
    if ((log->rotateCount == -1) && (log->rotateAge == 0))
        message(MESS_DEBUG, "old logs are kept forever\n");
    else if (log->rotateCount == 0)
        message(MESS_DEBUG, "no old logs will be kept\n");
    else {
        if (log->rotateCount == -1)
            message(MESS_DEBUG, "(unlimited rotations)");
        else
            message(MESS_DEBUG, "(%d rotations)", log->rotateCount);

        if (log->rotateAge > 0)
            message(MESS_DEBUG, ", old logs are removed after %d days\n", log->rotateAge);
        else if (log->logAddress)
            message(MESS_DEBUG, ", old logs mailed to %s\n", log->logAddress);
        else
            message(MESS_DEBUG, ", old logs are removed\n");
    }
```

kdudka (changes requested):
@cryptogopher I wonder how you tested the changes when the code fails to compile.  More comments inline...

logrotate.c
```diff
@@ -2372,8 +2378,6 @@ static int rotateLogSet(const struct logInfo *log, int force)
 
     if (log->logAddress) {
         message(MESS_DEBUG, "old logs mailed to %s\n", log->logAddress);
-    } else {
-        message(MESS_DEBUG, "old logs are removed\n");
```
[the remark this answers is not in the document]
cryptogopher: Right, missed this one. I guess it would be best to add newline outside of if statement.

logrotate.c
kdudka: From maintenance point of view, the number of `message()` calls does not really matter.  I primarily want to avoid fixing the same spelling mistakes (and copy/paste errors in the variadic arguments to `message()`) at multiple places in the code.  So the fewer duplications, the better.

As for old logs being either `mailed` or `removed`, it has always been presented this way.  You can see it as old logs being **moved** to mailbox rather than being removed.  It just matter of taste.

cryptogopher: After looking up the code I know that "mailed" means "moved to mail". But from user perspective "mailed" does not say whether this is "mailed while retaining copy on server" or "mailed and removed from server". Any wording that would remove ambiguity which - as you say - has been there since forever would be better than expecting users to know that "mailed" means "moved to mail".

As for the logic I really don't like when the same condition is spread in multiple places (e.g. `log->rotateCount == -1` in first and last `if`clause; the same for `log->rotateAge == 0`). This is in my opinion worse kind of duplication than duplicating debug messages and generating spelling mistakes. I understand that you're minimizing occurrences of particular message strings. But in my opinion it comes at too high cost of very hard to follow code logic and is error prone. I would rather have string duplication that hardly readable code.

kdudka: I think that documentation is clear in this:
```
      mail address
              When a log is rotated out of existence, it is mailed to address.
```

I have transformed my proposal to a regular pull request: #496

Issue #498, opened 2023-03-17
if prerotate return error, the log is not dumped and the log package will be renamed


banjiuqingshan:
![21DEC767-4A30-4484-92FA-651D17CFD0D0](https://user-images.githubusercontent.com/63209634/225863049-0c3bb852-438e-4495-b82e-be35a26dd151.png)
And we will lose the previous log

cgzones:
This is indeed the current behavior.
The loss of the previous log file in your examples is caused by overriding the log file via dd(1), otherwise it would have been preserved.
I am not sure what the historic reason is to rotate the old logs first, then run the prerotate script and then rotate the current log file; and I am also not sure if we should reorder it now.
Probably we could improve the documentation about this behavior:

https://github.com/logrotate/logrotate/blob/be6a263c87915d0467f5148d1488dd07f82be23a/logrotate.8.in#L667-L668

banjiuqingshan:
The original log will not be changed. I mean that logs that have been rotated will be renamed and overwritten, similar to test.log.3.gz, which will be renamed to test.log.4.gz, and then lost. In fact, all historical logs will be lost after several more executions. Can we reorder them or resume operations on the logs after the execution fails

banjiuqingshan:
The issue has been resolved by https://github.com/logrotate/logrotate/commit/fd1892e425f1facbbf1cee260e3bc5673a819182.

PR #502, opened 2023-03-22
resume logs if prerotate script fails

banjiuqingshan:
resume logs if prerotate script fails.#498


logrotate.c
```diff
@@ -2317,6 +2378,7 @@ static int rotateLogSet(const struct logInfo *log, int force)
     int numRotated = 0;
     struct logState **state;
     struct logNames **rotNames;
+    struct resumeConfigs *resCfgs;
```
kdudka, line 2381: The `resumeConfigs` struct can be allocated on the stack so that we do not need to handle its (de)allocation explicitly.

logrotate.c
```diff
@@ -2121,6 +2137,51 @@ static int prerotateSingleLog(const struct logInfo *log, unsigned logNum,
...
+{
+    int i;
+    char *newName = NULL;
+    char *oldName;
+    const char *fileext = resCfgs->fileext;
+    const char *compext = resCfgs->compext;
+    int rotateCount = resCfgs->rotateCount;
+    int logStart = resCfgs->logStart;
+
+    if (asprintf(&oldName, "%s/%s.%d%s%s", rotNames->dirName,
```
kdudka, line 2150: The code that constructs `oldname` should be moved to a helper function so that it is maintained at a single place.

kdudka (commented):
Before we dive into technical details ... @cgzones, could you please confirm that the new behavior is actually desired?

logrotate.c
banjiuqingshan: This place can indeed be refined. we can optimize it next time because it is also used elsewhere

cgzones:
I have no strong opinion on the behavior: the prerotate script failing is an error case, in which logrotate exists with a non 0 exit value and thus the system administrator should be notified (either via cron sending an email or the systemd timer entering the failing state). The man page currently only states

> If the scripts exit with error, the remaining actions will not be executed for the affected log only.

and

> The script is executed before the log file is rotated and only if the log will actually be rotated.

so the whether old logs are rotated on failure is not (for me at least) clearly stated.

---

Looking at the changed test cases isn't the change more about *reverting* the rotation of old logs than *resuming*?

banjiuqingshan:
I mean that logs that have been rotated will be renamed and overwritten, similar to test.log.3.gz, which will be renamed to test.log.4.gz.The commit will rename test.log.4gz back to test.log.3gz if the prerotate execution fails.Please consider whether this is necessary

cgzones:
This proposal does not *resume*/*restore* old logs of other files from the same directive when using `sharedscripts`.
Is this intened?
See #506 with the difference on *test-shared-pre-B.log*.

banjiuqingshan:
> This proposal does not _resume_/_restore_ old logs of other files from the same directive when using `sharedscripts`. Is this intened? See #506 with the difference on _test-shared-pre-B.log_.
This is a bug. I haven't had time to fix this recently. And being able to adjust the order of renaming would be better than resume logs.

cgzones:
Closing in favor of #506 (now merged).
Thanks for the patch nonetheless.

PR #506, opened 2023-04-04
do not rotate old logs on prerotate failure

cgzones:
Ensures old logs are preserved and not rotated out for logs with a failing prerotate script.

Alternative to #502

### DIFF
--- a/logrotate.8.in
+++ b/logrotate.8.in
@@ -702,8 +702,8 @@ the \fBSCRIPTS\fR section.
 \ \ \ \ \fIscript\fR
 .tq
 \fBendscript\fR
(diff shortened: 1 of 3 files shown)


kdudka (approved):
Should we also describe the behavior in the man page more precisely?

cgzones:
> Should we also describe the behavior in the man page more precisely?

Improved the man page a bit; or did you mean a complete action order list (*firstaction*, *prerotate*, *rotating old logs*, *rotating log file*, ..., *lastaction*)?

kdudka:
I think the updated wording is fine.  I admit I had to read it twice but hopefully understood in the end :-)

PR #508, opened 2023-04-23
Exit code equals 1 if configuration file is not owned by root

sblondon:
The [PR 448](https://github.com/logrotate/logrotate/pull/448) let me think the error messages strategy is:
MESS_ERROR if the error stops `logrotate`. So exit value should not be 0.
MESS_WARNING if it's possible to continue to run `logrotate`. Exit value is 0.

Currently, if configuration file is not owner by root, `logrotate` exits with 0 as exit code (it's in [confic.c](https://github.com/logrotate/logrotate/blob/master/config.c#L1070) ):

```C
message(MESS_ERROR,
    "Ignoring %s because the file owner is wrong (should be root or user with uid 0).\n",
    configFile);
```

```
$ sudo LANG=C ./logrotate logrotate.not_root.conf
error: Ignoring logrotate.not_root.conf because the file owner is wrong (should be root or user with uid 0).
$ echo $?
0
```

This PR changes the exit code to 1.

I found two cases where it should use `MESS_WARNING` because configuration lines are ignored:

1. [lines 342-343](https://github.com/logrotate/logrotate/blob/master/config.c#L342):
```C
message(MESS_ERROR, "%s:%d extra arguments for "
    "%s\n", configFile, lineNum, directive); 
```

```
$ sudo LANG=C ./logrotate logrotate.extra_argument.conf 
error: logrotate.extra_argument.conf:2 unknown option 'extraarg' -- ignoring line
$ echo $?
0
```

2. [lines 1746-1747](https://github.com/logrotate/logrotate/blob/master/config.c#L1746):
```C
message(MESS_ERROR, "%s:%d unknown option '%s' "
    "-- ignoring line\n", configFile, lineNum, key);
```

```
$ sudo LANG=C ./logrotate logrotate.invalid_directive.conf
error: logrotate.invalid_directive.conf:2 unknown option 'invaliddirective' -- ignoring line
$ echo $?
0
```

I can add them to the PR if you think it's the correct behavior.

Tests are all green. I didn't understand how to modify them to add exit value assertion and didn't find documentation about them. So, there is no additional test.

### DIFF
--- a/config.c
+++ b/config.c
@@ -1071,7 +1071,7 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
                     "Ignoring %s because the file owner is wrong (should be root or user with uid 0).\n",
                     configFile);
             close(fd);
-            return 0;
+            return 1;
         }
     }
 
--- a/config.c
+++ b/config.c
@@ -1743,7 +1743,7 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
                         message(MESS_DEBUG, "compress_ext is now %s\n",
                                 newlog->compress_ext);
                     } else {
-                        message(MESS_ERROR, "%s:%d unknown option '%s' "
+                        message(MESS_WARN, "%s:%d unknown option '%s' "
                                 "-- ignoring line\n", configFile, lineNum, key);
                         if (*start != '\n')
                             state = STATE_SKIP_LINE;
--- a/config.c
+++ b/config.c
@@ -1063,7 +1063,7 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
                     "Ignoring %s because it is writable by group or others.\n",
                     configFile);
             close(fd);
-            return 0;
+            return 1;
         }
 
         if (sb_config.st_uid != ROOT_UID) {


config.c
```diff
@@ -1071,7 +1071,7 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
                     "Ignoring %s because the file owner is wrong (should be root or user with uid 0).\n",
                     configFile);
             close(fd);
-            return 0;
+            return 1;
```
kdudka, line 1074: I think we should equally handle the `(sb_config.st_mode & 0022)` condition above.

kdudka (changes requested):
I think it makes sense to turn the other two errors into warnings, as you say.

As for checking logrotate's exit code in tests, you can have a look at these examples:
- https://github.com/logrotate/logrotate/blob/eb94f32dcc585e31c880418d82c38521076e882a/test/test-0108.sh#L18
- https://github.com/logrotate/logrotate/blob/eb94f32dcc585e31c880418d82c38521076e882a/test/test-0109.sh#L77

config.c
sblondon: I added it in new commit (https://github.com/logrotate/logrotate/pull/508/commits/86bffca5637a699354cb13d9d5cc7beb2d245910).

sblondon:
> Thanks for the update! As for the tests, one could instrument the call to `getuid()` at run-time. But let's not overcomplicate it.

Ok.
Do you want I add a line in `ChangeLog.md` file about this PR? Or it's good enough to be merged?

kdudka:
No worries.  I will update the change log in a batch before the next release.  I plan to merge this in a few days unless there is any more discussion.  Thanks for the contribution!

config.c
```diff
@@ -339,7 +339,7 @@ static int readModeUidGid(const char *configFile, int lineNum, const char *key,
     }
 
     if (rc == 4) {
-        message(MESS_ERROR, "%s:%d extra arguments for "
+        message(MESS_WARN, "%s:%d extra arguments for "
```
cgzones, line 342: This seems wrong to me: we are not ignoring the extra arguments but abort parsing them all altogether. And all callers check the return value and call `RAISE_ERROR()` if it is `-1`.

kdudka: @cgzones You are right.  Thank you for spotting it!

@sblondon Could you please revert this change?

kdudka: Thanks for the update!

kdudka (approved):

cgzones (approved):

Issue #514, opened 2023-05-04
maxage calculation is changed


banjiuqingshan:
With maxage set, any log file older than N days will to be purged.The relevant discussion can be seen in #39.

banjiuqingshan:
The use of difftime has caused this difference.
https://github.com/logrotate/logrotate/commit/4eb3add422f7d50128600eae3c678370fddc2217

kdudka:
I do not understand what exactly is wrong and what should be changed.

banjiuqingshan:
Before 4eb3add, when maxage was set to 1, files from 48 hours ago were not deleted and only deleted after 48 hours or more. Now, they will be deleted after 24 hours or more

kdudka:
Got it, thanks!  I have proposed pull request #515 to fix this.

PR #515, opened 2023-05-05
daysElapsed: use integer division for counting days

kdudka:
... to restore the original behavior of `logrotate-3.19.0`, before we started to use `difftime()`, which returns a floating point number.

Fixes: https://github.com/logrotate/logrotate/issues/514
Closes: https://github.com/logrotate/logrotate/pull/515


kdudka:
Obsoleted by #516.

PR #516, opened 2023-05-05
Restore old time_t difference calculations 

cgzones:
With the introduction of difftime(3) in commit https://github.com/logrotate/logrotate/commit/4eb3add422f7d50128600eae3c678370fddc2217 ("Use difftime
to compare time_t") to support platform with an unsigned time_t several
arithmetic operations and comparisons turned from integer to floating
point ones.  This can lead to different results since integer divisions
round down to the next integer while floating point divisions do not.

Cast the result of difftime(3) to intmax_t to revert back to integer
operations.

Fixes: https://github.com/logrotate/logrotate/commit/4eb3add422f7d50128600eae3c678370fddc2217 ("Use difftime to compare time_t")
Fixes: https://github.com/logrotate/logrotate/issues/514

Supersedes: #515

Additionally contains some test cases and minor changes.

### DIFF
--- a/test/test-common.sh
+++ b/test/test-common.sh
@@ -118,7 +118,7 @@ createlog() {
 	    what=seventh
 	    ;;
 	8)
-	    what=eight
+	    what=eighth
 	    ;;
 	9)
 	    what=ninth
--- a/test/test-common.sh
+++ b/test/test-common.sh
@@ -50,6 +50,19 @@ else
   exit 1
 fi
 
+# check for date(1) support of operating on a time given from the command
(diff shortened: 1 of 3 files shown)


test/test-common.sh
```diff
@@ -50,6 +50,17 @@ else
   exit 1
 fi
 
+if date --date @42 +%Y%m%d%H%M > /dev/null 2>&1; then
```
kdudka, line 53: Please put there a comment to explain what the code does and why we need it.

kdudka (approved):

PR #525, opened 2023-06-07
ci: add build job using C89 standard

cgzones:
Assure backwards compatibility with ancient compilers.

**TODO**: 
What minimum C standard do we actually want to support?
Currently it seems we support C89.


jamacku:
C89 seems like a reasonable version.

Issue #533, opened 2023-07-11
6 vulnerabilities in logrotate binary


blu3sh0rk:
<a name="xZDQN"></a>
# 6 vulnerabilities in logrotate binary

We found 6 vulnerabilities in logrotate binary and logrotate is complied with clang enabling ASAN.
```
$ git clone https://github.com/logrotate/logrotate

$ cd logrotate

$ CC=clang CXX=clang++ CFLAGS="-g -fsanitize=address -fno-omit-frame-pointer" CXXFLAGS="-g -fsanitize=address -fno-omit-frame-pointer" ./configure

$ make
```
<a name="xZDQN"></a>
# Reproduction
<a name="soAUT"></a>
#### Environment
OS: Ubuntu `20.04.5 LTS`
Compiler: `gcc version 9.4.0` or `clang-12`
version: commit `c9e3cb069b85dd57b6747d6f6b5da6039fae5421`
<a name="Rymwc"></a>
# 0x01 double-free in config.c:1846:31
```
$ ./logrotate -d poc1_double_free 
warning: logrotate in debug mode does nothing except printing debug messages!  Consider using verbose mode (-v) instead if this is not what you want.

reading config file poc1_double_free
error: cannot allocate memory [readConfigFile():1851]
=================================================================
==1305022==ERROR: AddressSanitizer: attempting double-free on 0x602000000650 in thread T0:
    #0 0x4988f9 in realloc (/home/blu3sh0rk/Fuzz/debug/logrotate/logrotate+0x4988f9)
    #1 0x4d6c11 in readConfigFile /home/blu3sh0rk/Fuzz/debug/logrotate/config.c:1846:31
    #2 0x4ca6eb in readConfigPath /home/blu3sh0rk/Fuzz/debug/logrotate/config.c:765:13
    #3 0x4c9a37 in readAllConfigPaths /home/blu3sh0rk/Fuzz/debug/logrotate/config.c:844:13
    #4 0x4dfdd8 in main /home/blu3sh0rk/Fuzz/debug/logrotate/logrotate.c:3239:9
    #5 0x7ffff7c0d082 in __libc_start_main /build/glibc-SzIz7B/glibc-2.31/csu/../csu/libc-start.c:308:16
    #6 0x41d71d in _start (/home/blu3sh0rk/Fuzz/debug/logrotate/logrotate+0x41d71d)

0x602000000650 is located 0 bytes inside of 8-byte region [0x602000000650,0x602000000658)
[...]

PR #534, opened 2023-07-12
Memory corruption issues on crafted configuration files

cgzones:
See individual commits and their descriptions fro details.

Closes: #533


jamacku (approved):
LGTM

The code looks good, but I haven't tested the reproducers. But I think this change shouldn't cause any regression, and it is ready to be merged.

Thank you, @cgzones, for the quick fix, and @blu3sh0rk, for reporting the issues.

PR #540, opened 2023-08-05
Use madvise(2) correctly

cgzones:
madvise(2) does only support one advice at a time, not multiple or'ed together, see the kernel implementation or their non-bitmask values.

Call madvise(2) for each advice individually.

Currently the result of MADV_SEQUENTIAL (=2), MADV_WILLNEED (=3) and MADV_DONTFORK (=10) results in MADV_DOFORK (=11).

See https://elixir.bootlin.com/linux/v6.4.8/source/mm/madvise.c#L1144


cgzones:
Maybe `MADV_DONTFORK` can be dropped entirely, since logrotate is single-threaded and during the lifetime of the memory mapping we are not going to fork.

kdudka:
Yes, I can see no evidence that `MADV_DONTFORK` would do anything useful in this case.

Issue #551, opened 2023-10-18
Dateformat %H doesn't work 


Forza-tng:
I am using logrotate-3.21.0 on Gentoo Linux AMD64.

The problem i am experiencing is that the `hour` is always `12`, no matter when logrotate runs. This is evident if we look at the timestamps from ls:

```
-rw------- 1 root root 40320853 Oct 18 14:02 bees.3t_backup.log
-rw------- 1 root root  3587829 Oct 17 21:06 bees.3t_backup.log-20231016-120700.zst
-rw------- 1 root root  2901724 Oct 18 07:01 bees.3t_backup.log-20231017-120111.zst
-rw------- 1 root root  6894067 Oct 18 13:00 bees.3t_backup.log-20231017-120112.zst
-rw------- 1 root root  4650221 Oct 18 03:01 bees.3t_backup.log-20231017-120123.zst
-rw------- 1 root root  3872284 Oct 18 00:33 bees.3t_backup.log-20231017-123328.zst
```

My logrotate.conf:

```
create
minsize 1M
maxsize 100M
notifempty
maxage 30
rotate -1
dateext
dateyesterday
dateformat -%Y%m%d-%H%M%S
compress
compresscmd /usr/bin/zstd
uncompresscmd /usr/bin/unzstd
compressext .zst
compressoptions --long --ultra -15 -T4
nomail

include /etc/logrotate.d
```

 /etc/logrotate.d/bees
```
/var/log/bees/*.log {
    copytruncate
    notifempty
    missingok
}
```

Logrotate is run via /etc/cron.hourly/logrotate
```bash
#!/bin/sh
echo `date` >> /var/log/logrotate.log
/usr/bin/logrotate /etc/logrotate.conf >> /var/log/logrotate.log 2>&1
EXITVALUE=$?
if [ $EXITVALUE != 0 ]; then
    /usr/bin/logger -t logrotate "ALERT exited abnormally with [$EXITVALUE]"
fi
echo `date` >> /var/log/logrotate.log
exit $EXITVALUE
```

cgzones:
Since you rotate hourly, you shouldn't use `dateyesterday` but `datehourago`.

Otherwise #554 removes the hardcoded noon time.

PR #554, opened 2023-10-19
Auto detect if DST in effect for dateyesterday

cgzones:
Instead of setting the hour unconditionally to 12 (noon), let mktime(3) detect whether DST is in effect at the specific time.

Closes: #551


cgzones:
Might regress on actual DST changes.
Maybe worth to wait for testing on the 29th (next DST change in Europe).

PR #561, opened 2023-11-12
Demonstrate mode/UID/GID parsing ambiguity

saper:
Since we accept numeric UID/GID, we are never
really sure which of the three optional
numeric parameters to "create" or "createolddir"
is really given.

Tried to update manual page to say which
parameters and optional and in which order.

### DIFF
--- a/test/Makefile.am
+++ b/test/Makefile.am
@@ -92,6 +92,7 @@ TEST_CASES = \
 	test-0091.sh \
 	test-0092.sh \
 	test-0093.sh \
+	test-0095.sh \
 	test-0100.sh \
 	test-0101.sh \
 	test-0102.sh \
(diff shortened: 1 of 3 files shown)


cgzones:
It seems historically `create foo bar` meant *default* mode, user *foo* and group *bar*, so for backward compatibility `create 1001 1002` should mean *default* mode, user *1001* and group *1002*.

Since the user and group strings are always converted to ids with regard to base 10 and the create mode normally should be specified as an octal value a leading `0` could also be used as hint whether the first token is the create mode or user id.

But I think these nuances can get very tricky so maybe it's best to come up with a new syntax altogether?

saper:
I think the test in this pull request does not pass, so the first two numbers (if the first one is parseable octally) become mode and the owner.

So I think we have a bug now since numeric UIDs have been introduced. We might need to fix it

Issue #567, opened 2023-12-11
systemd timer fuzz not implemented as intended


mschmitt:
In 2016, commit [05228c2](https://github.com/logrotate/logrotate/commit/05228c2f3cb39d39079c1c30ba998036850178f7) merged a systemd timer unit from OpenSuse, which was later amended by [3f3991e](https://github.com/logrotate/logrotate/commit/3f3991eedae3bb2f71a091ed43eef2392af9bcf0) to reduce "trigger fuzz".

I support a number of Linux workstation users who have enormous log volumes due to audit logging and noticed high load during system bootup. Investigation led towards [this systemd timer unit that can be traced back to logrotate's source repository](https://github.com/logrotate/logrotate/blob/096f92e4cbdc6811b68ccb5fffc590a094b9b6bb/examples/logrotate.timer):

```
# /lib/systemd/system/logrotate.timer
[Unit]
Description=Daily rotation of log files
Documentation=man:logrotate(8) man:logrotate.conf(5)

[Timer]
OnCalendar=daily
AccuracySec=1h
Persistent=true

[Install]
WantedBy=timers.target
```

The `Persistent=true` setting causes systemd to catch up with missed timers while the system was down. `AccuracySec=1h` does however not lead to a random trigger delay.

Quoting [the systemd documentation for systemd.timer](https://www.freedesktop.org/software/systemd/man/latest/systemd.timer.html#AccuracySec=):

> Note that this setting is primarily a power saving option that allows coalescing CPU wake-ups. It should not be confused with RandomizedDelaySec= (see below) which adds a random value to the time the timer shall elapse next and whose purpose is the opposite: to stretch elapsing of timer events over a longer period to reduce workload spikes.

My users will receive a packaged drop-in for the unit that amends this as follows:

```
# /usr/lib/systemd/system/logrotate.timer.d/randomize.conf
[Timer]
RandomizedDelaySec=1h
```

I'm submitting the issue in the hope that proper use of systemd timer randomization in the example unit is relevant to more people than just me. :-)

PR #574, opened 2024-01-05
Stretch systemd timer over one hour

cgzones:
The initial introduction of the systemd timer in commit 05228c2f ("Add logrotate service and timer for systemd") included the directive `AccuracySec=12h`.  AccuracySec= is not a load distribution option but a power saving one, see systemd-timer(5).  In commit 3f3991ee ("Timer unit: change trigger fuzz from 12h to 1h") the value was reduced to 1h, while the discussion in #230 also mentioned load distribution.

Use RandomizedDelaySec= instead of AccuracySec= to actually spread the start time of the service, and ignore potential system wakeups below the rate of the default of AccuracySec= of one minute.

Fixes: #567

### DIFF
--- a/examples/logrotate.timer
+++ b/examples/logrotate.timer
@@ -4,7 +4,7 @@ Documentation=man:logrotate(8) man:logrotate.conf(5)
 
 [Timer]
 OnCalendar=daily
-AccuracySec=1h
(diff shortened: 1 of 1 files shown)


jamacku (approved):
LGTM

PR #586, opened 2024-02-22
Deduplicate directories in a set

rmetrich:
When a directory is a symlink to another directory in the same set, logrotate was failing because it couldn't find the logs when processing the symlink, e.g. when the symlink was alphanumerically ordered after the directory it was referring to.
This patch removes this duplication through just skipping already known paths returned by globbing.

### DIFF
--- a/config.c
+++ b/config.c
@@ -1963,7 +1963,8 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
                             const struct logInfo *log;
                             struct stat sb_glob;
                             int add_file = 1;
-                            char *copy, *baseName, *dirName, *resolvedDir, *resolved_path;
+                            char *copy = NULL, *baseName = NULL, *dirName = NULL;
+                            char *resolvedDir = NULL, *resolved_path = NULL;
                             unsigned k;
 
                             /* if we glob directories we can get false matches */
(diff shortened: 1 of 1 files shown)


config.c
```diff
@@ -1911,19 +1911,78 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
...
+                            /* Resolves the directory but not the filename, to be able to detect symlinks */
+                            copy = strdup(globResult.gl_pathv[glob_count]);
+                            if (copy == NULL) {
+                                message_OOM();
+                                logerror = 1;
+                                goto duperror;
+                            }
+                            /* Because dirname() modifies the path, order is important */
+                            baseName = basename(copy);
+                            dirName = dirname(copy);
```
cgzones, line 1932: Both `basename()` and `dirname()` might modify their argument, so I don't think this is safe.

config.c
```diff
@@ -1911,19 +1911,78 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
...
+                                    goto duperror;
+                                }
+                            }
+                            resolved_path = (char *)malloc(strlen(resolvedDir) + 1 + strlen(baseName) + 1);
+                            if (resolved_path == NULL) {
+                                message_OOM();
+                                logerror = 1;
+                                goto duperror;
+                            }
+                            (void) sprintf(resolved_path, "%s/%s", resolvedDir, baseName);
```
cgzones, line 1963: use `asprintf()`

config.c
```diff
@@ -1911,19 +1911,78 @@ static int readConfigFile(const char *configFile, struct logInfo *defConfig)
...
+                            free(resolvedDir);
+                            free(copy);
+
+                            /* Skip duplicates for the current set */
+                            for (k = 0; k < newlog->numFiles; k++) {
+                                if (!strcmp(newlog->files[k], resolved_path)) {
+                                    message(MESS_DEBUG,
```
cgzones, line 1970: Is this worth `MESS_WARN`?

cgzones (commented):
Seems reasonable.
Please fix the memory leaks reported by the CI.

config.c
rmetrich: I don't think it worth `MESS_WARN` because it's not an issue as such.

rmetrich:
Hello, I agree there were memory issues, I reworked the code accordingly and used _valgrind_  to check new status.
I will check once CI runs again if any remains.

PR #590, opened 2024-03-14
Fix some RESOURCE_LEAKs

jamacku:
There is one leak in `setSecCtx()`, which this PR should resolve. The rest of the `freecon()` invocations are just to be safe.


cgzones:
The sanitizer failure is most likely caused by https://github.com/actions/runner-images/issues/9491 and fixed in [LLVM 17](https://github.com/llvm/llvm-project/commit/fb77ca05ffb4f8e666878f2f6718a9fb4d686839).

Please rebase.

logrotate.c
```diff
@@ -839,6 +840,9 @@ static int compressLogFile(const char *name, const struct logInfo *log, const st
 
     if (setSecCtx(inFile, name, &prevCtx) != 0) {
         /* error msg already printed */
+#ifdef WITH_SELINUX
+        freecon(prevCtx);
```
cgzones, line 844: Maybe move the cleanup into `setSecCtx()`?

jamacku: Thank you, that's a good point.

Please check if my patches don't do more harm than good. I am still new to the logrotate code.

logrotate.c
```diff
@@ -371,13 +371,16 @@ static int setSecCtx(int fdSrc, const char *src, char **pPrevCtx)
     /* save default security context for restoreSecCtx() */
     if (getfscreatecon_raw(pPrevCtx) < 0) {
         message(MESS_ERROR, "getting default context: %s\n", strerror(errno));
+        freecon(*pPrevCtx);
```
cgzones, line 374: I don't think we can free `*pPrevCtx` here, because callers still can use (free or restore from) it if SELinux is in permissve mode.

jamacku: Thank you, removed :+1:

Issue #105, opened 2017-01-30
Logrotate Errors on Username/Groups with Spaces


msangster:
(was originally submitted to https://bugzilla.redhat.com/show_bug.cgi?id=1417106)

# Description of problem:
If you use a space in a user/group name then logrotate will error out indicating that the usage was wrong.

# Version-Release number of selected component (if applicable):
I do not believe this is version related but in my test it was: logrotate-3.8.6-7.el7_2.x86_64

# How reproducible:
Every time

# Steps to Reproduce:
1. Create a logrotate configuration and use a group name with spaces in the "su" or "create" directives.
2. Run: `logrotate -d config`
3. Error output

# Actual results
```
cat /etc/logrotate.d/test
/var/log/test/*.log {
    su Test "Test Group"
    weekly
    rotate 10
    copytruncate
    delaycompress
    compress
    notifempty
    missingok
}

logrotate -d /etc/logrotate.d/test
reading config file /etc/logrotate.d/test
error: /etc/logrotate.d/test:2 extra arguments for su
error: found error in /var/log/test/*.log , skipping
removing last 1 log configs
Allocating hash table for state file, size 64 entries

Handling 0 logs
```

# Expected results:
Whilst spaces are not traditionally used in groups and users and indeed I am certain a lot of tools wouldn't accept it in the username. We have integration with Active Directory where it is acceptable. As such I would expect logrotate to handle the group name with spaces at least but preferably user and group.

# Additional info:
I did experiment with other possible styles of inputting the group.

su Test Test\ Group
error: /etc/logrotate.d/test:2 extra arguments for su

su Test "Test\ Group"
error: /etc/logrotate.d/test:2 extra arguments for su

su Test Test\040Group
error: /etc/logrotate.d/test:2 unknown group 'Test\040Group'

(1024 is the GID of "Test Group")
su Test 1024
error: /etc/logrotate.d/test:2 unknown group '1024'

alavrentiev:
You can try using this patch: https://github.com/logrotate/logrotate/issues/133
The plus sign (+) is used in place of spaces in group names.  User names cannot have spaces, though.
HTH

kdudka:
@alavrentiev In #133 you are proposing a one-line patch that fixes only a test-case whereas this bug is about logrotate not accepting user names and groups containing spaces while parsing the configuration files.

alavrentiev:
You can use a plus sign (instead of a space) in group names in config files, and with the patch they will work correctly.

[...]

PR #595, opened 2024-03-20
test: quote user and group names

cgzones:
Quote user and group names to avoid parsing failures on systems with spaces in the user or group name of the test user, e.g. CI systems.

Reported-by: alavrentiev

### DIFF
--- a/test/test-common.sh
+++ b/test/test-common.sh
@@ -95,7 +95,7 @@ genconfig() {
     user=$(id -u -n)
     group=$(id -g -n)
     rootgroup=$(id -g -n root)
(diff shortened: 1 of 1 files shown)


jamacku (approved):
LGTM

foxfire-auspex:
We're currently experiencing the same issue that was previously discussed in issue [#105](https://github.com/logrotate/logrotate/issues/105) as originally reported on [Red Hat's Bugzilla](https://bugzilla.redhat.com/show_bug.cgi?id=1417106).

The reason for some of our Linux systems' groups containing spaces is that they're derived from information we pull from Microsoft AD.

I'm just trying to assess the state and of and discussion surrounding this issue.

Sorry if I am misunderstanding, but it looks to me like this PR (an the cgzones:quote branch itself) only contains changes to tests.

Would this really fix logrotate's handling of group names containing spaces in a production environment (as @cgzones last comment under issue #105 suggests)?

foxfire-auspex:
Thank you very much for the clarification. I really appreciate you taking the time.

PR #604, opened 2024-05-24
ChangeLog.md: user-visible changes to appear in 3.22.0

cgzones:
Prepare a changelog entry for 3.22.0.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -6,6 +6,21 @@ All notable changes to this project will be documented in this file.
 
 [UNRELEASED]: https://github.com/logrotate/logrotate/compare/3.21.0...main
 
+## [3.22.0] - yyyy-mm-dd
+ - fix calculations for time differences (#516)
+ - fix extension for zip compression (#545)
+ - fix omitted copy for logs with `mail` and `rotate 0` (#553)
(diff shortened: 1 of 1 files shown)


kdudka (approved):
Looks good.  Please do not forget to:
- update the `UNRELEASED` link in `ChangeLog.md`
- create the `3.22.0` link in `ChangeLog.md`
- update the `latest release` and `previous releases` sections in `README.md`

cgzones:
I just released 3.22.0.
I would appreciate a short look if everything is updated correctly.

kdudka:
@cgzones Perfect, thank you for taking care of it!  I think we should update the [Verify and unpack](https://github.com/logrotate/logrotate?tab=readme-ov-file#verify-and-unpack) section of the top-level `README.md` file so that downstream consumers can easily verify signatures of the distribution tarballs.

kdudka:
> I think we should update the [Verify and unpack](https://github.com/logrotate/logrotate?tab=readme-ov-file#verify-and-unpack) section of the top-level README.md file so that downstream consumers can easily verify signatures of the distribution tarballs.

I can see this has already been reported as #611.

PR #612, opened 2024-06-08
Adjust file limit

cgzones:
The maximum number of files per directory is just a paranoid safety guard and probably one would even hit ENOMEM before. Adjust the limit a bit to avoid GCC stumbling on it when using LTO on 32-bit architectures:

    config.c:587:19: warning: iteration 1073741823 invokes undefined behavior [-Waggressive-loop-optimizations]
      587 |         free(array[i]);
          |                   ^
    config.c:586:19: note: within this loop
      586 |     for (i = 0; i < lines_count; ++i)
          |                   ^

### DIFF
--- a/config.c
+++ b/config.c
@@ -770,7 +770,9 @@ static int readConfigPath(const char *path, struct logInfo *defConfig)
         }
         while ((dp = readdir(dirp)) != NULL) {
             if (checkFile(dp->d_name)) {
-                if (files_count >= UINT_MAX - REALLOC_STEP) {
+                /* Use a lower limit than `UINT_MAX - REALLOC_STEP` to please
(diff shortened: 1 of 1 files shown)


config.c
```diff
@@ -770,7 +770,7 @@ static int readConfigPath(const char *path, struct logInfo *defConfig)
         }
         while ((dp = readdir(dirp)) != NULL) {
             if (checkFile(dp->d_name)) {
-                if (files_count >= UINT_MAX - REALLOC_STEP) {
+                if (files_count >= UINT_MAX / 4) {
```
kdudka, line 773: Could you please insert a comment to explain why `UINT_MAX / 4` is used?

kdudka (approved):

PR #616, opened 2024-06-20
Make wait() checks more reliable

Aparicio99:
**Reasoning** 

I've caught an issue where pre/post rotate scripts were being reported with an error "error running non-shared prerotate script for...", but only on very specific but deterministic conditions. After some debugging I've found it to be due to on my use case logrotate being launched from crond which had already inherited from its parent to ignore SIGCHLD.

Even in this situation, the issue depended on compiler optimization flags, and even on moving/deleting non related code around, which eventually found to be due to the uninitialized use of the wait() return argument, therefore the changes I propose for every use of wait()/waitpid().

---
**Commit message**

If wait()/waitpid() returns an error for some reason it will not write to the *wstatus argument. Checking wstatus in this case is non deterministic due to the variable not being initialized, and may lead to wrongly report the forked command as executed successfully.

Also, one possible reason for wait() to return an error even if the forked command has executed successfully is the child process exiting before we call wait(). Normally when the child process finishes it stays in a zombie state until the wait() call, but if our parent process has set to ignore the SIGCHLD signal, we will inherited it and our child processes will exit immediately. Restoring the signal handler on initialization will avoid this issue.

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -19,6 +19,7 @@
 #include <utime.h>
 #include <stdint.h>
 #include <libgen.h>
+#include <signal.h>
 
 #if !defined(PATH_MAX) && defined(__FreeBSD__)
 #include <sys/param.h>
@@ -516,9 +517,32 @@ static struct logState *findState(const char *fn)
     return p;
 }
 
-static int runScript(const struct logInfo *log, const char *logfn, const char *logrotfn, const char *script)
+static int waitpid_checked(pid_t pid, const char **errmsg)
 {
-    int rc;
+    int status;
+
+    if (waitpid(pid, &status, 0) < 0) {
+        *errmsg = strerror(errno);
+        return -1;
+    }
+
+    if (!WIFEXITED(status)) {
+        *errmsg = "child did not terminate normally";
+        return -2;
+    }
+
+    if (WEXITSTATUS(status) != 0) {
+        *errmsg = "child did not exit with status 0";
+        return -3;
+    }
+
+    *errmsg = NULL;
+    return 0;
+}
+
+static int runScript(const struct logInfo *log, const char *logfn,
+                     const char *logrotfn, const char *script, const char **errmsg)
+{
     pid_t pid;
 
     if (debug) {
@@ -545,8 +569,7 @@ static int runScript(const struct logInfo *log, const char *logfn, const char *l
         exit(1);
     }
 
-    wait(&rc);
-    return rc;
+    return waitpid_checked(pid, errmsg);
 }
 
 #ifdef WITH_ACL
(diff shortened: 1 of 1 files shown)


logrotate.c
```diff
@@ -3292,6 +3292,11 @@ int main(int argc, const char **argv)
 
     message(MESS_DEBUG, "\nHandling %d logs\n", numLogs);
 
+    /* Restore SIGCHLD handler in case our parent process had it ignored.
+     * Not doing this may cause our forked childs to exit prematurely before our
+     * wait() call, making it impossible to known the child proper exit code */
+    signal(SIGCHLD, SIG_DFL);
```
cgzones, line 3298: Please issue a warning on failure, e.g.:
```c
if (signal(SIGCHLD, SIG_DFL) == SIG_ERR)
    message(MESS_WARN, "failed to reset SIGCHLD handler: %s\n", strerror(errno));
```

cgzones (changes requested):
Thanks, looks good to me except one small nit.

logrotate.c
Aparicio99: Updated with your suggestion.

logrotate.c
```diff
@@ -545,8 +546,11 @@ static int runScript(const struct logInfo *log, const char *logfn, const char *l
         exit(1);
     }
 
-    wait(&rc);
-    return rc;
+    if (wait(&rc) > 0 && WIFEXITED(rc)) {
```
cgzones, line 549: Could you please take this opportunity and replace with `wait(3)` with `waitpid(3)`.
```c
waitpid(pid, &rc, 0);
```

logrotate.c
```diff
@@ -753,9 +757,7 @@ static int shred_file(int fd, const char *filename, const struct logInfo *log)
         exit(1);
     }
 
-    wait(&status);
-
-    if (!WIFEXITED(status) || WEXITSTATUS(status)) {
+    if (wait(&status) < 0 || !WIFEXITED(status) || WEXITSTATUS(status)) {
```
cgzones, line 760:
```c
waitpid(pid, &status, 0);
```

logrotate.c
```diff
@@ -989,19 +991,19 @@ static int compressLogFile(const char *name, const struct logInfo *log, const st
     }
 
     close(compressPipe[0]);
-    wait(&status);
 
-    fsync(outFile);
-
-    if (!WIFEXITED(status) || WEXITSTATUS(status)) {
+    if (wait(&status) < 0 || !WIFEXITED(status) || WEXITSTATUS(status)) {
```
cgzones, line 995:
```c
waitpid(pid, &status, 0);
```

cgzones (commented):
I'd like to take this opportunity and replace the `wait(3)` calls with `waitpid(3)`.
Also the `waitpid(3)` call on line 1112 is not checked.

Aparicio99:
Replaced all wait() calls with waitpid() and added validation of all return codes.

strerror(errno) can also be added to the error messages, but since it would only apply to when waitpid() fails and not when "!WIFEXITED(status) || WEXITSTATUS(status)" is true, it would complicate the code and possible have 2 different error messages for each call. Not sure if it's worth it, but I could also add this if you think so.

cgzones:
> Replaced all wait() calls with waitpid() and added validation of all return codes.

Thanks.

> strerror(errno) can also be added to the error messages, but since it would only apply to when waitpid() fails and not when "!WIFEXITED(status) || WEXITSTATUS(status)" is true, it would complicate the code and possible have 2 different error messages for each call. Not sure if it's worth it, but I could also add this if you think so.

What about a helper like:

```c
int waitpid_checked(pid_t pid, const char **errmsg) {
    int r, status;

    r = waitpid(pid, &status, 0);
    if (r <= 0) {
        *errmsg = strerror(errno);
        return -1;
    }

    if (!WIFEXITED(status)) {
        *errmsg = "child did not terminate normally";
        return -1;
    }

    if (WEXITSTATUS(status) != 0) {
        *errmsg = "child did not exit with status 0";
        return -1;
    }

    *errmsg = NULL;
    return 0;
}
```
?

Aparicio99:
Implemented the helper as you suggested. Also updated runScript and callers to use the helper error message.

cgzones (approved):

cgzones:
Applied. Thanks for your contribution and the patch iterations.

Issue #579, opened 2024-01-22
stay silent about firstaction error


uli42:
We have a process that can run for 1-2 days and want to rotate its logfiles usign logrotate. However, the logfiles MUST NOT be rotated while that process is running and it cannot be signalled or stopped intermediately. So a colleague implemented a config file similar to this:
```
/var/log/logfile
{
  sharedscripts
  firstaction
    # abort logrotate if some-process is running
    ! /bin/pgrep some-process >/dev/null
  endscript
  copytruncate
  missingok
  notifempty
  rotate 30
  compress
}

So the idea is to check for the process in question and return an error if it is running to make logrotate skip this partiular rotation.

This works, but everytime the process is found running firstaction will return an error and logroate will inform us via mail (from cron) like this:
```
/etc/cron.daily/logrotate:

error: error running first action script for /var/log/logfile
```

As there's no need to send a mail in that case we are wondering if there's a way to silently (= no mail) skip rotation? If not consider this a feature request ;-)

Degoah:
Would be nice to a kind of option to keep logrotate silent. Look for the same feature.

cgzones:
Please give #624 a try.

PR #624, opened 2024-07-27
Introduce checkaction

cgzones:
Add a script directive that runs similar to firstaction before the whole log group, but on script failure skips gracefully the rotations instead of reporting the failure.

Closes: #579


cgzones:
Another name would be `conditionaction`.

PR #626, opened 2024-07-27
Add support for non-existent uid/gid

cgzones:


cwrx777:
thanks for adding this feature.

PR #627, opened 2024-07-28
Add more dateformat specifiers

ryancdotorg:
Adds support for the following specifiers to `dateformat`:

* `%G` - ISO 8601 week-based year
* `%y` - two digit version of `%Y`
* `%g` -  two digit version of `%G`
* `%U` - week number variant
* `%W` - week number variant
* `%u` - day of week as 1 to 7, Monday being 1
* `%w` - day of week as 0 to 6, Sunday being 0
* `%j` -  day of year as 001 to 366

The two digit year variants break lexical sorting when year 2100 rolls around, but `%s` is already supported and has the same problem in the 23rd century. They're convenient for now, but if you don't want to leave this particular footgun around for our ancestors to find, I can remove them. If it's 2100 and you're reading this after your logs broke, you should have updated your config. 🤷

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [UNRELEASED]
 
 [UNRELEASED]: https://github.com/logrotate/logrotate/compare/3.22.0...main
+ - Add support for %G, %y, %g, %U, %W, %u, %w, and %j to dateformat. [ryancdotorg]
 
 ## [3.22.0] - 2024-06-01
  - fix calculations for time differences (#516)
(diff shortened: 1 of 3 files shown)


ryancdotorg:
Also, my math says `%s` hits 11 digits on 2286-11-20?

cgzones:
> Also, my math says `%s` hits 11 digits on 2286-11-20?

Seems correct:

```
$ date --date='@10000000000'
Sat Nov 20 06:46:40 PM CET 2286
```

cgzones (approved):

cgzones:
Merged, thanks for your contribution.

Issue #632, opened 2024-08-02
`test_0112.sh` fails on SELinux-enabled system


fishilico:
Hello,
On a Arch Linux system running SELinux, `make check` fails:

```console
$ make check
[...]
PASS: test-0111.sh
FAIL: test-0112.sh
============================================================================
Testsuite summary for logrotate 3.22.0
============================================================================
# TOTAL: 107
# PASS:  106
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
```

Here are the test logs:

```console
$ cat test/test-0112.sh.log
Running test 112
No error printed, but there should be one.
FAIL test-0112.sh (exit status: 3)
```

```console
$ cat test/test-0112/error.log 
reading config file test-config.112
Creating stub state file: state
acquired lock on state file state
Reading state from file: state
Allocating hash table for state file, size 64 entries

Handling 1 logs

rotating pattern: /home/vagrant/.tmp/build/logrotate-selinux/src/logrotate-3.22.0/test/test-0112/test*.log forced from command line empty log files are rotated, (2 rotations), old logs are removed
considering log /home/vagrant/.tmp/build/logrotate-selinux/src/logrotate-3.22.0/test/test-0112/test_fifo.log
Creating new state
  Now: 2024-08-02 11:51
  Last rotated at 2024-08-02 11:00
  log needs rotating
considering log /home/vagrant/.tmp/build/logrotate-selinux/src/logrotate-3.22.0/test/test-0112/test_reg.log
Creating new state
  Now: 2024-08-02 11:51
  Last rotated at 2024-08-02 11:00
  log needs rotating
rotating log /home/vagrant/.tmp/build/logrotate-selinux/src/logrotate-3.22.0/test/test-0112/test_fifo.log, log->rotateCount is 2
dateext suffix '-20240802'
glob pattern '-[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]'
error: error opening /home/vagrant/.tmp/build/logrotate-selinux/src/logrotate-3.22.0/test/test-0112/test_fifo.log: Operation not supported
rotating log /home/vagrant/.tmp/build/logrotate-selinux/src/logrotate-3.22.0/test/test-0112/test_reg.log, log->rotateCount is 2
dateext suffix '-20240802'
glob pattern '-[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]'
set default create context to sysadm_u:object_r:user_home_t
[...]

PR #633, opened 2024-08-03
Avoid opening log file for getting SELinux context

cgzones:
Currently setSecCtxByName() uses open_logfile() to get a file descriptor to the current log file to retrieve its security context. open_logfile() performs additional checks, like whether the file is a regular file, which alter the control flow between systems with SELinux enabled and disabled.  This can be observed in the reported issue #632. Use lgetfilecon_raw() instead to have the same behavior for SELinux enabled and disabled systems and delay the checks for invalid log files to code executed in both cases.

Closes: #632

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -361,7 +361,34 @@ static int movefd(int oldfd, int newfd)
     return rc;
 }
 
-static int setSecCtx(int fdSrc, const char *src, char **pPrevCtx)
+#ifdef WITH_SELINUX
+static int setSecCtx(char *srcCtx, char **pPrevCtx)
+{
+    /* save default security context for restoreSecCtx() */
+    if (getfscreatecon_raw(pPrevCtx) < 0) {
+        message(MESS_ERROR, "getting default context: %s\n", strerror(errno));
+        freecon(srcCtx);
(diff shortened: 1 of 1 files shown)


fishilico (approved):
This change looks good to me. Thanks for working on this :)

logrotate.c
```diff
@@ -409,29 +409,51 @@ static int setSecCtx(int fdSrc, const char *src, char **pPrevCtx)
...
+        message(MESS_ERROR, "setting default context to %s: %s\n", srcCtx,
+                strerror(errno));
+        freecon(*pPrevCtx);
+        *pPrevCtx = NULL;
+        freecon(srcCtx);
+        return selinux_enforce;
+    }
+
+    message(MESS_DEBUG, "set default create context to %s\n", srcCtx);
+    freecon(srcCtx);
```
kdudka, line 451: This code is copy/pasted from `setSecCtx()`.  It would be better to move it to a new function that would be called by both `setSecCtx()` and `setSecCtxByName()` so that it does not need to be maintained at two places.

cgzones: good idea, updated.

kdudka (approved):

PR #645, opened 2024-11-12
doc: Add Security Policy

jamacku:


jamacku:
From what I know, we don't have a mailing list. However, I think that the GitHub feature for reporting security issues might be enough. Do you have any opinions, @cgzones?

cgzones:
No, there is no mailing list that I am aware of.

The GitHub security advisory page is as far as I tested only accessible for project members, not regular users.

Issue #637, opened 2024-09-15
Monthly frequency should have monthday specifier the same way week has a weekday specifier


eusousu:
Man logrotate(8) states:
weekly [weekday]
              Log  files  are  rotated once each weekday, or if the date is advanced by at least 7
              days since the last rotation (while ignoring the exact time).  The weekday interpre‐
              tation is following: 0 means Sunday, 1 means Monday, ..., 6 means Saturday; the spe‐
              cial value 7 means each 7 days, irrespectively of weekday.  Defaults  to  0  if  the
              weekday argument is omitted.
              
I propose that monthly should have a similar specifier that maps the month day like so:
monthly [monthday]
              Log  files are rotated each monthday, or the first time logrotate is run in a month 
              (this is normally on the first day of the month). Monthdays are numbered from 1-31
              and, if a month would not have an specified day (i.e. your set monthday is 31 and 
              it's February) the rotation happens on the earliest available date (in the example,
              either 28 or 29, depending if it's a leap year)

PR #660, opened 2025-04-11
Add support for monthday

cgzones:
Add support for rotating on a specific day of the month.

Closes: #637


kdudka (approved):
Looks good.  Some test coverage (at least by appending a number to the `monthly` directive in an existing test case) would be nice.

Issue #581, opened 2024-01-25
Feature Request: Set default statefile via configuration


hjf288:
It would be helpful to be able to specify the default state file name/location via the configuration file rather than have it defined at compile time or passed via command line.

cgzones:
What is your use case?
Normally the default state file fits the default config files run by root, and if specified a separate config file one can also specify a separate state file.

hjf288:
Sorry missed the notification for this.

I'm using log rotate in a ephemeral container and want to ensure that state is persisted.

I've modified the daily cron to define the location to the state file that I want, however when invoking `logrotate` via the CLI (arguably a rare situation) then I can't guarantee that someone is going to specify the location to the state file.

I can't use a symlink for `/var/lib/logrotate/status` as the symlink gets overwritten with a file - so I'd like to just be able to set the state file to the location it needs to be via configuration file.

You mention that a separate state file can be set via config - how is this done? I don't see reference to it in the man page, only passing it as a parameter to the `log rotate` command.

cgzones:
> I can't use a symlink for /var/lib/logrotate/status as the symlink gets overwritten with a file - so I'd like to just be able to set the state file to the location it needs to be via configuration file.

This seems like a current oversight; since commit https://github.com/logrotate/logrotate/commit/3e7b5e4fc1a48348908f0f5154347d736b7be281 the symlink should not be overridden anymore, but probably the state should be written to `readlink(2)` of the state file instead.

PR #664, opened 2025-04-11
Handle state file symlink

cgzones:
In case the state file is a symlink override the target and create the temporary file in the targets parent directory instead of the symlink itself.

Only follows one level of symlink.

Impreoves: #581


kdudka (changes requested):
I am afraid this would introduce a security weakness in case the symlink (or the directory where it resides) is write-accessible by a non-privileged user.  If `logrotate` runs with root privileges, a non-privileged user would be able to overwrite files they do not have otherwise write access to.

PR #668, opened 2025-05-15
config:  use correct .v in default taboo list

prownd:
In the default taboo ext array, the wildcard suffix `,v` in the configuration file should be spelled incorrectly,
The correct suffix should be `.v`, which is the suffix of the configuration file or backup file.

Additionally, update the description of `.v` in the man manual

### DIFF
--- a/config.c
+++ b/config.c
@@ -140,7 +140,7 @@ enum {
 };
 
 static const char *const defTabooExts[] = {
-    ",v",
+    ".v",
     ".bak",
     ".cfsaved",
     ".disabled",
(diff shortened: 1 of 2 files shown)


kdudka (changes requested):
@prownd Thanks for heads up!  The spelling though seems to be correct as it is.  The `,v` suffix seems to be used by the RCS file revision control system: https://pages.cs.wisc.edu/~plonka/sysadmin/article.html

Could you please clarify where you get files with the `.v` suffix?

prownd:
> @prownd Thanks for heads up! The spelling though seems to be correct as it is. The `,v` suffix seems to be used by the RCS file revision control system: https://pages.cs.wisc.edu/~plonka/sysadmin/article.html
> 
> Could you please clarify where you get files with the `.v` suffix?

Oh, yes, according to your description and documentation, it seems to be correct, but the `,v` file suffix is strange and cannot be understood.
I noticed that there is a `.v` file suffix, but after searching and analyzing, I found that there are several file types, but now it doesn't seem to be related to configuration backup.

prownd:
So, close it

PR #675, opened 2025-08-07
Add check to verify that temp status file from previous logrotate run is present before attempting to delete it

willisman31:
In environments where syscalls are tightly monitored, attempting to delete a file which is not present will result in an audit report of a failed syscall despite logrotate working as it should; this check simply aims to reduce auditing noise

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -2798,8 +2798,10 @@ static int writeState(const char *stateFilename)
     }
     strcpy(tmpFilename, stateFilename);
     strcat(tmpFilename, ".tmp");
-    /* Remove possible tmp state file from previous run */
-    error = unlink(tmpFilename);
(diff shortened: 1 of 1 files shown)


logrotate.c
```diff
@@ -2798,8 +2798,10 @@ static int writeState(const char *stateFilename)
     }
     strcpy(tmpFilename, stateFilename);
     strcat(tmpFilename, ".tmp");
-    /* Remove possible tmp state file from previous run */
-    error = unlink(tmpFilename);
+    /* If present, remove possible tmp state file from previous run */
```
kdudka, line 2801: On the other hand, this introduces [TOCTOU](https://cwe.mitre.org/data/definitions/367.html), which will trigger static analyzers.  It is not an improvement really.

Issue #700, opened 2026-03-05
AIX build failure due to incompatible struct types


shubhamhii:
While building logrotate in AIX, we are facing build issues, below is the log

```

gcc -DHAVE_CONFIG_H -I.  -include config.h  -Wall -Wconversion -Wdeclaration-after-statement -Wextra -Wmissing-format-attribute -Wmissing-noreturn -Wmissing-prototypes -Wpointer-arith -Wshadow -Wstrict-prototypes -Wundef -Wunused -Wwrite-strings -g -O2 -MT logrotate.o -MD -MP -MF .deps/logrotate.Tpo -c -o logrotate.o logrotate.c
logrotate.c: In function 'setAtimeMtime':
logrotate.c:805:13: error: incompatible types when assigning to type 'struct timespec' from type 'st_timespec_t' {aka 'const struct st_timespec'}
  805 |     ts[0] = sb->st_atim;
      |             ^~
logrotate.c:806:13: error: incompatible types when assigning to type 'struct timespec' from type 'st_timespec_t' {aka 'const struct st_timespec'}
  806 |     ts[1] = sb->st_mtim;
  logrotate.c: At top level:
logrotate.c:1155: warning: "MIN" redefined
 1155 | #define MIN(a,b) ((a) < (b) ? (a) : (b))
      | 
In file included from /usr/include/sys/file.h:29,
                 from logrotate.c:11:
/opt/freeware/lib/gcc/powerpc-ibm-aix7.1.0.0/10/include-fixed/sys/param.h:183: note: this is the location of the previous definition
  183 | #define MIN(a,b) (((a)<(b))?(a):(b))
      | 
make[2]: *** [Makefile:531: logrotate.o] Error 1
make[2]: Leaving directory '/home/buildusr/rpmbuild/BUILD/logrotate-3.21.0/32bit
```

PR #701, opened 2026-03-06
fix for copying incompatible struct timespec from st_timepsec_t in AIX

shubhamhii:
#700 

Reason for fix:
In AIX, 
 since the structs(struct timespec ts and struct st_timespec_t st_atim, st_mtime) are same by layout, memcpy can be used instead of assignment operator for copying data between two structures solving the problem of incompatible types.


cgzones:
I prefer a type-safe alternative, like #708

shubhamhii:
> I prefer a type-safe alternative, like #708

@cgzones 
Thanks, that works for me.

cgzones:
Superseded by #708

PR #703, opened 2026-04-30
sparse: fix incorrect detection due to mismatched block size units

### LINKED ISSUE #682: LogRotate incorrectly detects file sparseness
LogRotate's [is_probably_sparse()](https://github.com/logrotate/logrotate/blob/main/logrotate.c#L1179-L1188) function fails to detect sparse files due to mismatched block size units. While the system reports blocks in 512B through st_blocks, the function compares against the filesystem block size (st_blksize). This leads to incorrect sparse file detection on filesystem that use a value larger than 512B block size.

Using the copytruncate configuration, this issue can cause regions of null bytes to be written to disk as actual data, resulting in unnecessary disk space usage.

AndyLau-SOC:
## Summary
- Fix `is_probably_sparse()` function in `logrotate.c` that incorrectly compared `st_blocks` (512-byte units per POSIX) with `st_blksize` (filesystem preferred I/O size)
- When `st_blksize` differs from 512, sparse files on such filesystems could be incorrectly detected as non-sparse (or vice versa)
- This caused `copytruncate` to expand holes into actual data, wasting disk space
- Remove the now-unnecessary `HAVE_STRUCT_STAT_ST_BLKSIZE` preprocessor check and `st_blksize != 0` guard

Fixes #682

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## [UNRELEASED]
 
+ - fix incorrect sparse file detection due to mismatched st_blocks and st_blksize units (#682)
+
 [UNRELEASED]: https://github.com/logrotate/logrotate/compare/3.22.0...main
  - Add support for %G, %y, %g, %U, %W, %u, %w, and %j to dateformat. [ryancdotorg]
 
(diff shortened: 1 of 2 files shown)


cgzones (approved):

cgzones:
Applied.
Thanks for your contribution.

PR #710, opened 2026-05-26
fix(copytruncate): do not skip copy when postrotate script exists

### LINKED ISSUE #667: Postrotate data-loss regression with copytruncate and rotate 0
The change to resolve #191 introduced a regression in the behaviour of postrotate scripts in existing configurations, resulting in data loss.

Prior to #191, because `copytruncate` unconditionally performed the copy even with `rotate 0`, the postrotate scripts would be able to see the `logfile.1` temporary file _before_ it was then deleted, eg. enabling them to copy the data to a central archive.

After the change, because of the optimisation to skip the copy when `rotate 0` is in force, postrotate scripts no longer have a chance to process the about-to-be-deleted file.

I suggest _not_ setting the `skip_copy` flag whenever there are postrotate scripts that would be run for the file in question, rather than making it depend solely on the value of `rotate`.

I think this is reasonable, because if the administrator genuinely doesn't care about the data, then they are unlikely to have defined a postrotate script at all.  `copytruncate` only exists for the case where the writing process doesn't support reopening its logs, so it would be rare for there to be a postrotate script that just does a `kill -HUP` and does _not_ look at the data file.  At worst, this would merely add a few edge cases where the copy happens when it wasn't strictly needed.

Issue #443 also raised this issue.  The workaround mentioned there (to add an explicit `rotate 1`) is valid - but it is only a workaround, and in some cases would change the behaviour (eg. if the postrotate script copies rather than moves the first-generation copy, then logrotate won't delete it either, and it will remain on disc).

(I know #191 is old, but the affected versions of logrotate only started appearing in RHEL9 distributions, so the regression has only more recently become widely visible.)

AndyLau-SOC:
When rotate 0 is used with copytruncate, the copy step was skipped
unconditionally in rotateSingleLog(), causing postrotate scripts to
lose access to the log data before deletion.

This change preserves the copy when a postrotate script is defined
(log->post != NULL), since the script is likely to need the data
being rotated.

Also update the corresponding skipped_copy check in
postrotateSingleLog() for consistency.

Fixes: #667

### DIFF
--- a/logrotate.c
+++ b/logrotate.c
@@ -2377,7 +2377,7 @@ static int rotateSingleLog(const struct logInfo *log, unsigned logNum,
                 && !(log->flags & LOG_FLAG_TMPFILENAME)) {
             hasErrors = copyTruncate(log->files[logNum], rotNames->finalName,
                                      &state->sb, log,
-                                     !log->rotateCount && !log->logAddress);
+                                     !log->rotateCount && !log->logAddress && !log->post);
         }
 
 #ifdef WITH_ACL
@@ -2422,7 +2422,8 @@ static int postrotateSingleLog(const struct logInfo *log, unsigned logNum,
         int skipped_copy = (log->flags & (LOG_FLAG_COPYTRUNCATE | LOG_FLAG_COPY)) &&
                            !(log->flags & LOG_FLAG_TMPFILENAME) &&
                            !log->rotateCount &&
-                           !log->logAddress;
+                           !log->logAddress &&
+                           !log->post;
 
         if (!skipped_copy)
             hasErrors = compressLogFile(rotNames->finalName, log, &state->sb);


cgzones (approved):

logrotate.c
```diff
@@ -2421,7 +2422,7 @@ static int postrotateSingleLog(const struct logInfo *log, unsigned logNum,
         int skipped_copy = (log->flags & (LOG_FLAG_COPYTRUNCATE | LOG_FLAG_COPY)) &&
                            !(log->flags & LOG_FLAG_TMPFILENAME) &&
                            !log->rotateCount &&
-                           !log->logAddress;
+                           !log->logAddress && !log->post;
```
cgzones: nit: put `!log->post` into a new line.

AndyLau-SOC: Done, thanks for the review.

cgzones (approved):

PR #723, opened 2026-08-25
config: propagate and release addextension in log configurations

prownd:
copyLogInfo() copied every string member of struct logInfo except addextension, and freeLogInfo() released every string member except addextension.  This caused two defects:

  * addextension specified in the global (default) section was silently dropped for every log file definition that followed it, because each definition is created by newLogInfo() -> copyLogInfo() from the default configuration.  Every other filename directive (extension, compressext, olddir, ...) is inherited, so this was inconsistent with both the documentation and user expectation.

  * the string was leaked, which is reported by LeakSanitizer.

Reproducer for the first defect:

    $ cat lr.conf addextension .wow /tmp/logs/app.log.wow { daily rotate 3 } $ logrotate -d -f lr.conf | grep 'renaming.*app.log.wow ' renaming /tmp/logs/app.log.wow to /tmp/logs/app.log.wow.1

The same configuration with "addextension .wow" moved inside the braces correctly renames to /tmp/logs/app.log.1.wow.

Reproducer for the leak:

    Direct leak of 5 byte(s) in 1 object(s) allocated from:
        #1 in isolateLine config.c:198
        #2 in isolateValue config.c:230
        #3 in readConfigFile config.c:1808

Copy and free addextension alongside extension.  Add test 114, which fails before this change and passes after it.


kdudka (approved):
@prownd Looks good. Thanks!

PR #724, opened 2026-08-25
Report the real popt error for a bad command line option

prownd:
main() passes 'rc' to poptStrerror() when reporting a bad command line option, but the popt error code is in 'arg'; 'rc' is still 0 at that point, so every command line error was reported as "unknown error":

    $ logrotate --bogus-option x
    logrotate: bad argument --bogus-option: unknown error
    $ logrotate --state
    logrotate: bad argument --state: unknown error

With the fix, popt's actual diagnostic is shown:

    logrotate: bad argument --bogus-option: unknown option
    logrotate: bad argument --state: missing argument


prownd:
@kdudka Please review it.

kdudka (approved):
@prownd Looks good. Thanks!

PR #725, opened 2026-08-25
Name the file that is actually in the way when rotation is skipped

prownd:
prerotateSingleLog() stats destFile but names rotNames->firstRotated in the "destination already exists" error message.  With dateext, compress and delaycompress the two differ by the compression extension, so the message named a file that does not exist:

    error: destination /var/log/app.log-20260824 already exists,
    skipping rotation

while the file actually in the way was app.log-20260824.gz.  Report destFile, the name that was actually tested.


prownd:
@kdudka Please review it.

kdudka (approved):
@prownd Looks good.  Thanks!

PR #726, opened 2026-08-25
config: keep the line counter in sync in parseGlobString()

prownd:
parseGlobString() consumes the newlines of a log file definition that spans several lines, but the caller's line counter is passed by value, so those lines are never counted.  Every diagnostic emitted after such a definition reports a wrong line number.  With this config:

     1  # line 1
     2  /var/log/a.log
     3  /var/log/b.log
     4  /var/log/c.log
     5  {
     6      missingok
     7  }
     8
     9  # this bad directive is on line 10
    10  bogusdirective 1

logrotate reported "lr.conf:7 unknown option 'bogusdirective'".  It now correctly reports lr.conf:10.  Pass the counter by pointer and increment it on every newline consumed.


prownd:
@kdudka Please review it

kdudka:
@prownd Looks reasonable, thanks!  Could you please cover the fix by a test-case?

prownd:
> @prownd Looks reasonable, thanks! Could you please cover the fix by a test-case?

You're right, a new test case should be added.

kdudka (approved):
@prownd Thanks for the update!  Works as expected.

